Issue #1, opened 2016-12-14
Improve Documentation


tippenein:
The Haddock documentation is [pretty good](https://hackage.haskell.org/package/esqueleto-2.4.3/docs/Database-Esqueleto.html) but the readme should include the most useful bits of these examples.

It should highlight `where_`, `sub_select`, `InnerJoin`, etc. It would also be nice to have input on common "gotcha's" and import practices in order to avoid overlap with Persistent (`import qualified Database.Esqueleto as E`).

bitemyapp:
I actually just mentioned the Persistent/Aeson overlap on Slack. That drives me crazy. I think docs are a good short-term fix since I'd like to keep changes minimal while we get caught up on everything.

bitemyapp:
I think this is resolved by the PR, but we should fork a couple issues from this. I'll reference this issue in them, file a more specific issue if something isn't covered.

PR #2, opened 2016-12-14
README port from Haddocks

tippenein:
This can provide a starting point for improving the initial impression of Esqueleto.

There are slight changes from the [Haddock](https://hackage.haskell.org/package/esqueleto-2.4.3/docs/Database-Esqueleto.html) but ordering and examples have been retained.

Addresses #1


bitemyapp:
@tippenein this is fine except for the manual intra-graf linebreaks. I can fix this post-merge or you can. What do you prefer?

bitemyapp:
@tippenein I'll take care of it.

PR #9, opened 2017-01-07
Use correct `in_` quoting in haddock code block

qrilka:
Follow-up on https://github.com/prowdsponsor/esqueleto/pull/154


bitemyapp:
Thank you again :)

PR #10, opened 2017-01-07
Correct persistent module name for example imports

qrilka:
Follow-up to https://github.com/prowdsponsor/esqueleto/pull/152


bitemyapp:
Thank you!

PR #18, opened 2017-02-04
Fix postgres tests

ChrisCoffey:
I've added back some missing imports that were preventing the postgres tests from building.

I've also updated the `travis.yml` to configure a local postgres install & run the tests against postgres.


bitemyapp:
@ChrisCoffey Thank you very much!

PR #20, opened 2017-02-12
use persistent-sqlite >= 2.1.3

ncaq:
persistent use persistent-sqlite >= 2.1
if remove support old persistent-sqlite, we can remove ugly preprocessor.
and, text editor can detect symmetry of blacket.


bitemyapp:
This should in principle be okay, but would it be alright if I sit on this until I cut the next release?

ncaq:
Yes, This PR is not high priority.
Think you.

PR #26, opened 2017-04-15
Allow persistent 2.7.0

psibi:


psibi:
It would be nice if a new version of esqueleto is released with this change.

Issue #29, opened 2017-06-21
Is there a way to reuse a query executed via selectSource?


dashea:
I'm having an issue using `selectSource` (sqlite backend). Since unlike `select` the SQL statement is still running instead of the results being all pulled into memory, the stmtReset finalizer isn't being called. Reusing the `Source` results in a ErrorMisuse from sqlite, even if I use it in separate conduits, since the statement is only prepared once and there is no `sqlite3_reset` call between the executions.

Is there a way that the finalizer actions for selectSource could be called sooner (or explictly?), or a way to have it prepare separate statements instead of reusing the same one? Or is there something I need to be doing with the MonadResource?

Thanks for any help or advice.

Here's an example, two conduits with a selectSource source, error on the second one:
```haskell
{-# LANGUAGE GADTs #-}
{-# LANGUAGE GeneralizedNewtypeDeriving #-}
{-# LANGUAGE MultiParamTypeClasses #-}
{-# LANGUAGE OverloadedStrings #-}
{-# LANGUAGE QuasiQuotes #-}
{-# LANGUAGE TemplateHaskell #-}
{-# LANGUAGE TypeFamilies #-}

import Control.Monad(void)
import Control.Monad.IO.Class(liftIO)
import Control.Monad.Trans.Resource(MonadResource)
import Data.Conduit(($$), (=$=), Source)
import qualified Data.Conduit.List as CL
import Database.Esqueleto
import Database.Persist.Sqlite(runSqlite)
import Database.Persist.TH

share [mkPersist sqlSettings, mkMigrate "migrateAll"] [persistLowerCase|
Person
    name String
    deriving Show
|]

selectPerson :: MonadResource m => String -> Source (SqlPersistT m) (Key Person)
selectPerson name = do
    let source = selectSource $ from $ \person -> do
                       where_ $ person ^. PersonName ==. val name
                       return $ person ^. PersonId
    source =$= CL.map unValue

main :: IO ()
main = runSqlite ":memory:" $ do
    runMigration migrateAll

    void $ insert $ Person "John Smith"
    void $ insert $ Person "Jane Doe"

    selectPerson "John Smith" $$ CL.consume >>= liftIO . print
    selectPerson "Jane Doe"   $$ CL.consume >>= liftIO . print
```

bitemyapp:
I'm investigating the issue here: https://github.com/bitemyapp/esqueleto-select-source-error-misuse

[...]

PR #30, opened 2017-06-21
release key with selectSource

parsonsmatt:
The current implementation of `selectSource` does not call the release key, instead waiting for the scope to end with `runResourceT`.

With the SQLite backend, this causes a runtime exception as reported in #29.

This PR introduces a test case for the problem and a solution.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,6 +32,7 @@ script:
   - stack build
   - stack test --flag esqueleto:postgresql
   - stack test --flag esqueleto:-mysql
+  - stack test
 
 cache:
   directories:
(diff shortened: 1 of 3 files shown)


bitemyapp:
@parsonsmatt could you do me a favor and add an undecorated `stack test` invocation so that it runs the `sqlite` tests in TravisCI? Currently it seems like it only runs `MySQL` and `PostgreSQL`.

parsonsmatt:
done~

bitemyapp:
@parsonsmatt I'm going to call this a SemVer `patch` bump as it's a backwards compatible bug fix. Thank you!

PR #43, opened 2017-07-31
Cleaning up Esqueleto errors

### LINKED ISSUE #37: Specific error types
When `esqueleto` errors, it `error`s. It seems like a very nice quick fix would be to purely throw a real exception type with context rather than stringly typed errors.

bitemyapp:
Fixes #37


bitemyapp:
@FintanH I stupidly mangled your PR by accidentally pulling in `mysql-rand` which isn't supposed to be part of `master`, it's just a fix-up branch for my workplace. I used cherry-pick to create a clean branch for this PR.

PR #45, opened 2017-08-03
Replace data with newtype in Value and ValueList.

greydot:
From what I can see, the linked issue (#6124) was fixed somewhere along the way, and I don't get any warnings from compiling the testcase from the ticket with ghc-7.10 and later versions.

### DIFF
--- a/src/Database/Esqueleto/Internal/Language.hs
+++ b/src/Database/Esqueleto/Internal/Language.hs
@@ -16,7 +16,6 @@ module Database.Esqueleto.Internal.Language
     Esqueleto(..)
   , from
   , Value(..)
-  , unValue
   , ValueList(..)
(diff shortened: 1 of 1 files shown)


bitemyapp:
Could you link the issue you're talking about please?

bitemyapp:
Ah, okay, I think you were referring to this GHC ticket: https://ghc.haskell.org/trac/ghc/ticket/6124

Please let me know if I was wrong.

greydot:
Correct. I should have been more clear on this perhaps.

bitemyapp:
@greydot S'no problem at all. I'll review this later tonight. I don't think anything will block the merge but I'd like to grok the GHC ticket first.

Thank you!

PR #46, opened 2017-08-03
Now

FintanH:


bitemyapp:
@FintanH https://travis-ci.org/bitemyapp/esqueleto/builds/260756675?utm_source=github_status&utm_medium=notification couple things to fix in the build, one of them dependency lower bound, another the imports.

Even after you fix those, I think for the tests to be correct and passing you'd want to wrap this test stanza for `now_` in the PostgreSQL-only cpp macro right?

esqueleto.cabal
```diff
@@ -102,7 +102,7 @@ test-suite test
     , persistent-template >= 2.1
     , monad-control
     , monad-logger        >= 0.3
-    , time                >= 1.6.0.1
+    , time                >= 1.5.0.1 && <= 1.6.0.1
```
bitemyapp: higher upper bound than that please, need to test it with 1.8.0.1

FintanH: Ah ok will do 👍

bitemyapp:
Thank you @FintanH!

PR #48, opened 2017-08-10
Remove cpp

FintanH:
Removed the CPP pragmas. The tactic for this was to isolate all common tests under `Common/Test.hs` and split the tests for running into different stanzas: postgresql, sqlite, mysql.

At the moment there can be some code duplication, say if one test works the same on both mysql and sqlite but not postgresql they still go in the separate Test.hs files. 

Each of the separate modules defines the backend it's running with, which also ends up looking like duplicate code.

But at least everything looks a lot cleaner and is separated out into digestible parts.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,9 +30,9 @@ script:
   - stack setup
   - stack update
   - stack build
-  - stack test --flag esqueleto:postgresql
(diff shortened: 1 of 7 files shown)


bitemyapp:
@FintanH merged this, thank you very much!

PR #53, opened 2017-09-13
Add a class for backend compatibility

parsonsmatt:
This PR will enable Esqueleto queries to be used by the `persistent-typed-db` library, along with other database backends.

This is a WIP, please don't merge until I've had time to use it in-anger.

### DIFF
--- a/esqueleto.cabal
+++ b/esqueleto.cabal
@@ -66,14 +66,14 @@ library
       base                 >= 4.8     && < 5.0
     , bytestring
     , text                 >= 0.11    && < 1.3
-    , persistent           >= 2.7.1   && < 2.8
+    , persistent           >= 2.8.0   && < 2.9
     , transformers         >= 0.2
     , unordered-containers >= 0.2
     , tagged               >= 0.2
 
     , monad-logger
-    , conduit              >= 1.1
-    , resourcet            >= 1.1
+    , conduit              >= 1.3
+    , resourcet            >= 1.2
     , time                 >= 1.5.0.1 && <= 1.8.0.2
     , blaze-html
   hs-source-dirs: src/
--- a/esqueleto.cabal
+++ b/esqueleto.cabal
@@ -66,7 +66,7 @@ library
       base                 >= 4.8     && < 5.0
     , bytestring
     , text                 >= 0.11    && < 1.3
-    , persistent           >= 2.5     && < 2.8
+    , persistent           >= 2.7.1   && < 2.8
     , transformers         >= 0.2
     , unordered-containers >= 0.2
     , tagged               >= 0.2
--- a/esqueleto.cabal
+++ b/esqueleto.cabal
@@ -98,7 +98,7 @@ test-suite test
     , HUnit
     , QuickCheck
     , hspec               >= 1.8
-    , persistent-sqlite   >= 2.1.3
+    , persistent-sqlite   >= 2.8.0
     , persistent-template >= 2.1
     , monad-control
     , monad-logger        >= 0.3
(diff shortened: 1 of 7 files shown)


parsonsmatt:
This PR now uses the backend compatibility class introduced in `persistent-2.7.1` instead of defining it here.

The work project "just worked" with this patch and the new persistent, so I suspect it's fine. I'll try to find someone with heavier usage of the `SqlReadT` stuff to see if it is OK for them.

parsonsmatt:
I've opened [this PR](https://github.com/yesodweb/persistent/pull/723) in `persistent` that should make this a non-breaking change for existing projects. Once that's reviewed/merged, I'll bump the lower bounds on this library, use the hackage version, and we'll be in business :)

dariooddenino:
Using both this PR and the persistent one you linked, I still get:

```
    • No instance for (BackendCompatible SqlBackend (SqlFor LivDB))
        arising from a use of ‘E.select’
    • In a stmt of a 'do' block:
        E.select
        $ E.from
          $ \ p
              -> do { E.where_ (p E.^. Wp_postmetaMeta_id E.==. E.val x);
                      return p }
      In the second argument of ‘($)’, namely
        ‘do { E.select $ E.from $ \ p -> do { ... } }’
      In the expression:
        runLiv $ do { E.select $ E.from $ \ p -> do { ... } }

```

It works with standard persistent queries. It's a mysql database (using persistent-mysql-2.6.1)

dariooddenino:
Using both this PR and the persistent one you linked, I still get this error:

```
 • No instance for (BackendCompatible SqlBackend (SqlFor LivDB))
        arising from a use of ‘E.select’
    • In a stmt of a 'do' block:
        E.select
        $ E.from
          $ \ p
              -> do { E.where_ (p E.^. Wp_postmetaMeta_id E.==. E.val x);
                      return p }
      In the second argument of ‘($)’, namely
        ‘do { E.select $ E.from $ \ p -> do { ... } }’
      In the expression:
        runLiv $ do { E.select $ E.from $ \ p -> do { ... } }
```

Persistent queries work correctly. It's a mysql db on `persistent-mysql-2.7.1`

bitemyapp:
@dariooddenino please post a more complete repro of the problem.

parsonsmatt:
@dariooddenino Try with latest commit of `persitent-typed-db` -- I forgot to move the instance into the library :)

dariooddenino:
`persistent-typed-db` last commit fixed this :)

dariooddenino:
Does this works for `select` only? If I try to use `update` I get the `Couldn't match type 'SqlFor Foo' with 'SqlBackend' error.
Let me know if you need more info :)

parsonsmatt:
@dariooddenino I have added tests for the `persistent-typed-db` library that ensure that it type checks with updates and selects now :)

@bitemyapp This is working and relying on Hackage `persistent`. However, upstream `persistent` bundles along new Conduit and a bunch of other breaking changes. Merging and releasing this would be a major bump probably. Thoughts?

bitemyapp:
@parsonsmatt I lean towards doing major bumps whenever in doubt. I will have time this weekend to help with this too.

parsonsmatt:
Cool. If you're good with a major bump for this, then Im happy with that too. Anything we want to get in with this version?

bitemyapp:
@parsonsmatt other than any relevant persistent and conduit upgrades? Not that I'm aware of off-hand but I'll re-peruse the diff and be more certain after I've done so. You got anything in mind?

parsonsmatt:
Idk, there's probably some deprecations or stuff that can make the interface better?

bitemyapp:
@parsonsmatt I have some work to do yet on cleaning up the test suite but this is now merged. Thank you very much!

PR #60, opened 2017-10-13
Fix SQL syntax in join example

mheinzel:
I'm pretty new to SQL, but I think this should be fixed.


bitemyapp:
@mheinzel thanks!

Issue #61, opened 2017-10-17
Add helper for projecting nullable fields


bigs:
hey!

not sure if this is wanted, but in a project i'm working on i had need to construct expressions with values that could be null, guarding against their nullity with a `WHERE` clause. i've forked the project and contributed the functions to the `Esqueleto` class. if there's any interest in having this merge upstream, i'll add some tests and make a pr! thoughts?

bitemyapp:
I've merged https://github.com/bitemyapp/esqueleto/pull/62

I'm interested in the class you're suggesting if you'd still like to upstream that.

PR #62, opened 2017-10-17
Add withNonNull helper to project nullable values

bigs:
Guards against null values with a where_ call. Includes a test as well.

Addresses #61


bigs:
in a local branch i've actually defined a 

```haskell
class WithNonNull a b where
    withNonNull :: a -> (b -> SqlQuery c) -> SqlQuery c
```

with instances for the base case as well as some recursive instances for tuples so you can do e.g.

```haskell
withNonNull (m1, m2, m3) $ \(m1', m2', m3') -> do
  -- ...
```

is this of any interest?

Vlix:
Haven't had the need for it (yet), but seems like a nice convenience function.

bitemyapp:
@bigs thank you!

PR #66, opened 2018-01-10
Add arrayRemove wrapper for PostgreSQL function

sestrella:
Add support for `array_remove` function.

https://www.postgresql.org/docs/10/static/functions-array.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,8 @@ addons:
       - postgresql-client
       - postgresql-server-dev-all
 
+  postgresql: "9.6"
+
 env:
  - GHCVER=7.10
  - GHCVER=8.0
--- a/src/Database/Esqueleto/PostgreSQL.hs
+++ b/src/Database/Esqueleto/PostgreSQL.hs
@@ -6,6 +6,7 @@
 module Database.Esqueleto.PostgreSQL
   ( arrayAggDistinct
   , arrayAgg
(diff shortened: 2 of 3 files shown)


bitemyapp:
Looks fine in principle, need to resolve the merge conflict (that's you) and figure out why the test failed (either of us, I'll be taking a look as it wasn't obvious why it failed)

sestrella:
@bitemyapp yes I also wasn't sure about what the build fail after taking a look at the log in travis since `stack test` output code was 0. I'm going to solve the conflicts right away.

bitemyapp:
Think I found the source of the error: 

screenshot from 2018-01-10 12-20-05

sestrella:
@bitemyapp thank you I'm going to take a look at that.

.travis.yml
```diff
@@ -13,6 +13,8 @@ addons:
       - postgresql-client
       - postgresql-server-dev-all
 
+  postgresql: "9.6"
```
sestrella: @bitemyapp I bumped PostgreSQL version. Hopefully, that solves the issue since I tested locally using the same version `9.6`.

sestrella:
@bitemyapp bumping PostgreSQL did the trick. Thanks for your help.

bitemyapp:
@sestrella thanks for the contributions and the follow-up!

PR #75, opened 2018-03-01
Various additions

Philonous:
Over time I have written a bunch of addons to esqueleto to make my life easier and I couldn't imagine living without them now. So I've packaged the entire smörgåsbord up into this PR because I don't know what parts of it might be suitable for inclusion. If there's interest I'll move the relevant parts into their own PRs.

# Overview of Contents:

## 1. Foreign Key handlers
A system for handling foreign key references (see [ForeignKey.hs](https://github.com/Philonous/esqueleto/blob/foreignkeys/src/Database/Esqueleto/ForeignKey.hs)). The idea here is that foreign key relationships can be captured with the ForeignKey class.

So you instead of 

```haskell
  from $ \(team `InnerJoin` employee) ->
    on (team ^. TeamEmployee ==. Employee ^. EmployeeNum
        &&. <...>
        )
    <...>
```
After defining 

```haskell
instance ForeignKey Team Employee where
  keyPairs = [KeyPair TeamEmployee EmployeeNum
              -- possibly more KeyPairs in case of composite keys
             ]
```
you can write 

```haskell
  from $ \(team `InnerJoin` employee) ->
    onForeignKey team employee
    <...>
```

which saves a lot of typing, especially on composite keys

There's also some TH code that generates the `ForeignKey` instances automatically from the persistent schema definion (just add `mkForeignInstances` to the `share` invocation). This of course only works for entity pairs with exactly one foreign reference.

Please note that this adds a new dependency on template-haskell. However, most people using esqueleto are probably already using persistent-template, so it's unclear how onerous this new dependency is.

## 2. Folds 

`andL`, `orL`, `whereL`, `onL` and their respective `*Mb` variants fold over lists (resp. lists of Maybes). so instead of chains of `&&.` and `||.` you can write lists, which is IMO cleaner, especially in conjuction with list comprehension.

example: 
```haskell 
where_ $ orL [ x ^. y >. foo | Left foo <- foos] 
```

## 3. various PostgreSQL-specific functions

See changes to [PostgreSQL.hs](https://github.com/Philonous/esqueleto/blob/foreignkeys/src/Database/Esqueleto/PostgreSQL.hs)

### DIFF
--- a/src/Database/Esqueleto/ForeignKey.hs
+++ b/src/Database/Esqueleto/ForeignKey.hs
@@ -1,0 +1,245 @@
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TupleSections #-}
+
+module Database.Esqueleto.ForeignKey
+  ( ForeignPair
+  , ForeignKey
+  , mkForeignInstances
+  , onForeignKey
+  , foreignKey
+  , foreignKeyL
+  , foreignKeyR
+  , foreignKeyLR
+  , foreignKeyLMaybe
+  , foreignKeyRMaybe
+  , foreignKeyLRMaybe
+  )
+
+where
+
+import qualified Data.Char                                    as Char
+import qualified Data.Function                                as Function
+import qualified Data.List                                    as List
+import           Data.Monoid
+import qualified Data.Ord                                     as Ord
+import qualified Data.Text                                    as Text
+import           Data.Traversable                             (forM)
+import qualified Language.Haskell.TH                          as TH
+
+import           Database.Esqueleto.Internal.Language
+import           Database.Esqueleto.Internal.PersistentImport
+
+-- | Describe a Pair of keys that are part of a foreign key relationship.
+--
+-- The first element is the entity field that holds the foreign key. The second
+-- element is the entity field that holds the references primary key. Type of
+-- primary key and foreign key have to coincide.
+--
+-- For example, give the following Entity definition:
+--
+-- @
+-- Employee
+--     num Int
+--     Primary numid
+--     name Text
+--
+-- Team
+--     teamId Int
+--     employee Int
+--     Foreign Employee fkEmployee employee
+-- @
+--
+-- The Following ForeignPair would capture the foreign key relationship
+--
+-- @ForeignPair TeamEmployee EmployeeNum@
+
+data ForeignPair :: * -> * -> * where
+    ForeignPair :: forall a b f.
+                   (PersistEntity a, PersistEntity b, PersistField f) =>
+                    EntityField a f
+                 -> EntityField b f
+                 -> ForeignPair a b
+
+-- | Describe a unique, canonical foreign key relationship between entities. For
+-- example, given the entity definitions from 'ForeignPair', there is exactly
+-- one foreign key relationship between Employee and Team, so we can capture it
+-- in a type class:
+--
+-- @
+-- instance ForeignKey Team Employee where
+--     foreignPairs = [ForeignPair TeamEmployee EmployeeNum]
+-- @
+--
+-- Note that the entity with the foreign key is the _first_ parameter of the
+-- type class, the target entity the second
+class ForeignKey a b where
+  foreignPairs :: [ForeignPair a b]
+
+-- | Apply f to each pair of foreign fields (most likely some variant of equality)
+withForeignPairs ::
+     (ForeignKey a b, Esqueleto query expr backend)
+  => (forall f. (PersistField f, PersistEntity a, PersistEntity b)  =>
+                EntityField a f
+             -> EntityField b f
+             -> expr (Value Bool))
+  -> expr (Value Bool)
+withForeignPairs f = andL . flip map foreignPairs $ \case
+  (ForeignPair xk yk) -> f xk yk
+
+-- | A foreign key constraint between two entities.
+--
+-- Example:
+--
+-- @
+-- from $ \(team, employee) ->
+--   where_ (foreignKey team employee)
+--   [...]
+-- @
+foreignKey :: (ForeignKey a b, Esqueleto query expr backend) =>
+              expr (Entity a) -> expr (Entity b) -> expr (Value Bool)
+foreignKey x y = withForeignPairs $ \xk yk -> x ^. xk ==. y ^. yk
+
+-- | 'foreignKey' for 'RightOuterJoin'
+foreignKeyR  :: (ForeignKey a b, Esqueleto query expr backend) =>
+               expr (Entity a) -> expr (Maybe (Entity b)) -> expr (Value Bool)
+foreignKeyR x y = withForeignPairs $ \xk yk ->just (x ^. xk) ==. y ?. yk
+
+-- | 'foreignKey' for 'LeftOuterJoin'
+foreignKeyL  :: (ForeignKey a b, Esqueleto query expr backend) =>
+               expr (Maybe (Entity a)) -> expr (Entity b) -> expr (Value Bool)
+foreignKeyL x y = withForeignPairs $ \xk yk ->(x ?. xk) ==. just (y ^. yk)
+
+-- | 'foreignKey' for 'FullOuterJoin'
+foreignKeyLR  :: (ForeignKey a b, Esqueleto query expr backend) =>
+               expr (Maybe (Entity a)) -> expr (Maybe (Entity b)) -> expr (Value Bool)
+foreignKeyLR x y = withForeignPairs $ \xk yk ->(x ?. xk) ==. (y ?. yk)
+
+-- | Like foreignKeyL, but also matches if the foreign reference is NULL
+foreignKeyLMaybe :: (Esqueleto query expr backend, ForeignKey a b) =>
+                    expr (Maybe( Entity a))
+                 -> expr (Entity b)
+                 -> expr (Value Bool)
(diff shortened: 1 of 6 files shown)


src/Database/Esqueleto/ForeignKey.hs
```diff
@@ -0,0 +1,245 @@
+{-# LANGUAGE GADTs #-}
+{-# LANGUAGE KindSignatures #-}
+{-# LANGUAGE LambdaCase #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE RankNTypes #-}
+{-# LANGUAGE TemplateHaskell #-}
+{-# LANGUAGE TupleSections #-}
+
+module Database.Esqueleto.ForeignKey
```
parsonsmatt: IMO: This module seems really useful, but also might be better suited as a library in addition to Esqueleto. That would allow you to iterate much quicker on the design of it without having to worry about breaking existing Esqueleto features.

src/Database/Esqueleto/Internal/Language.hs
```diff
@@ -1060,3 +1072,59 @@ instance ( Esqueleto query expr backend
...
+----------------------------------------------------------------------
+-- Functions that Work on lists --------------------------------------
+----------------------------------------------------------------------
+
+-- | @OR@ a list of predicates. An empty list becomes 'False'
+orL :: Esqueleto query expr backend =>
+       [expr (Value Bool)]
+    -> expr (Value Bool)
+orL [] = val False
+orL (p:ps) = List.foldl' (||.) p ps
```
parsonsmatt: This can just be `orL = List.foldl' (||.) (val False)`

src/Database/Esqueleto/Internal/Language.hs
```diff
@@ -1060,3 +1072,59 @@ instance ( Esqueleto query expr backend
...
+
+-- | @AND@ a list of predicates. An Empty list becomes 'True'
+andL :: Esqueleto query expr backend =>
+        [expr (Value Bool)] -> expr (Value Bool)
+andL [] = val True
+andL (p:ps) = List.foldl' (&&.) p ps
```
parsonsmatt: See the comment on `orL`  -- is there a specific reason you wrote it like this?

src/Database/Esqueleto/Internal/Language.hs
```diff
@@ -1060,3 +1072,59 @@ instance ( Esqueleto query expr backend
...
+
+-- | Set offset and limit for the query.
+offsetLimit  :: (Esqueleto m expr backend ) =>
+                Maybe Int
+             -> Maybe Int
+             -> m ()
+offsetLimit os l = do
+    Foldable.forM_ os $ offset . fromIntegral
+    Foldable.forM_ l $ limit . fromIntegral
+    return ()
```
parsonsmatt: nit: the `return ()` is redundant, as `forM_` has an `m ()` return already.

src/Database/Esqueleto/Internal/Language.hs
```diff
@@ -32,6 +32,15 @@ module Database.Esqueleto.Internal.Language
...
   , ToBaseId(..)
+    -- ** Functions that work on lists
+  , orL
+  , andL
+  , andLMb
+  , whereL
+  , whereLMb
+  , onL
+  , onLMb
+  , offsetLimit
```
parsonsmatt: I'm a little dubious on the inclusion of these functions, to be honest -- they're all somewhat trivial compositions of preexisting functions. `foldl' (&&.) (val True)` and `foldl' (||.) (val False)` are definitely things I've written in my own code before, but they don't seem quite involved enough to warrant inclusion into the library.

src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -41,13 +53,26 @@ random_ = unsafeSqlValue "RANDOM()"
 arrayAgg :: SqlExpr (Value a) -> SqlExpr (Value [a])
 arrayAgg = unsafeSqlFunction "array_agg"
 
+-- | Like 'array_agg' but removes @NULL@ values
+arrayAgg' :: PersistField [a] => SqlExpr (Value (Maybe a)) -> SqlExpr (Value [a])
+arrayAgg' =  arrayRemoveNull . arrayAgg
```
parsonsmatt: A `'` by itself typically means "this value, but modified in some way," where the modification is obvious in the code you're reading. A more self-documenting name (eg `arrayAggNonNull`) would be clearer.

src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -41,13 +53,26 @@ random_ = unsafeSqlValue "RANDOM()"
...
 --
 -- /Since: 2.5.3/
 arrayRemove :: SqlExpr (Value [a]) -> SqlExpr (Value a) -> SqlExpr (Value [a])
 arrayRemove arr elem' = unsafeSqlFunction "array_remove" (arr, elem')
 
+-- | Remove @NULL@ values from an array
+arrayRemoveNull :: SqlExpr (Value [Maybe a]) -> SqlExpr (Value [a])
+arrayRemoveNull x = unsafeSqlFunction "array_remove" (x, unsafeSqlValue "NULL")
+
+
```
parsonsmatt: nit: extra whitespace line here

src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -1,22 +1,30 @@
...
   , arrayAgg
+  , arrayAgg'
   , arrayRemove
+  , arrayRemoveNull
+  , array
   , stringAgg
   , chr
   , now_
   , random_
+  , toChar
```
parsonsmatt: The additions in this module look great! Would you be willing to write some tests for the `unsafeSqlFunction` bits?

src/Database/Esqueleto/Internal/Language.hs
Philonous: Yes, they should be identical, but removing the superfluous False results in simpler SQL which makes debugging it easier (especially if you have a bunch of them in a query - which I tend to have).

parsonsmatt: Ah, that makes sense! Might want to leave it as a comment in the code, then :slightly_smiling_face:

src/Database/Esqueleto/ForeignKey.hs
```diff
@@ -0,0 +1,245 @@
...
+-- @ForeignPair TeamEmployee EmployeeNum@
+
+data ForeignPair :: * -> * -> * where
+    ForeignPair :: forall a b f.
+                   (PersistEntity a, PersistEntity b, PersistField f) =>
+                    EntityField a f
+                 -> EntityField b f
+                 -> ForeignPair a b
+
+-- | Describe a unique, canonical foreign key relationship between entities. For
```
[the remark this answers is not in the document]
Philonous: Yes, that restriction is pretty unfortunate. An alternative would be to create named values instead and refer to them. The cost is that now we have to remember the name of the foreign-key relationship (instead of the name of the fields), which the typeclass approach avoids. Persistent does indeed store the foreign key information in various forms, but at least as far as I can see only as `Text`s.

src/Database/Esqueleto/Internal/Language.hs
Philonous: The *Mb-variants are rather trivial, I agree.

src/Database/Esqueleto/PostgreSQL.hs
bitemyapp: @parsonsmatt I tend to agree. I'm okay with prime-markers (apostrophe) as signifying distaff or strictness, but not anything to do with `Maybe`.

src/Database/Esqueleto/ForeignKey.hs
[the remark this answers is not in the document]
bitemyapp: >The cost is that now we have to remember the name of the foreign-key relationship

If I am not mistaken, isn't this the case with how SQL database work anyway? You have to name your unique constraints in the support Persistent offers for this already, so it's not a departure from how Persistent works either.

It seems like you could generate sum type of valid foreign key relationships tied to a model to have well-typed references to FKs that aren't stringly. Similar to Yesod routing perhaps if you're familiar with that. I don't know what the Persistent representation looks like, that might be more suitable.

bitemyapp:
I really appreciate you upstreaming this so I want to try to honor your contributions by being as detailed as I can.

# Foreign Key

I have two thoughts on this, one is that this probably needs more time to bake before it should be upstreamed and could-should function best as an independent library. I don't think it's necessary that it live in Esqueleto at this time but I _do_ think you should publish it and iterate on the design.

>This of course only works for entity pairs with exactly one foreign reference.

This is a disconcerting limitation for something I'd be considering for inclusion in Esqueleto.

I think it would be a shame if you didn't publish this feature in particular because I wish people felt more free to make their own quasiquoters for the purpose of extending the schema language.

# Folds

This is a hair on the wrong side of "people can write this themselves." Another candidate for "maybe in a library" but I don't know if it'd be worth publishing an extra package apart from the foreign relationships stuff. Naturally that's up to you.

I'm including the `*Mb` functions more generally in this but I do appreciate the first-time user experience w/ `Maybe`/left joins and `just`, `isNothing`, `==.`, and `?.` is far from ideal.

# Various PostgreSQL-specific functions

I'm pretty interested in getting this merged, possibly via separate PR or a rebase of this one at your preference. However, I have been fixing some issues that arose partly due to a lack of test coverage the last couple days. As a result, I need to be more disciplined about requiring tests for PR acceptance from the original contributor that I'm not guessing at how something is meant to work after the fact. With that in mind, would you please add tests exercising each new PostgreSQL-specific function?

Philonous:
Thank you for looking over this. I will clean up and add tests to the PostgreSQL functionality and move it into it's own PR.

Philonous:
OK, the parts that were mergeable are in #76, so I'll close this.

bitemyapp:
@Philonous thank you again!

PR #76, opened 2018-03-08
Postgresql additions

### LINKED ISSUE #68: stringAgg can return null, but the type doesn't indicate that
The type of `stringAgg` is

```haskell
stringAgg 
  :: SqlString s
  => SqlExpr (Value s)
  -> SqlExpr (Value s)
  -> SqlExpr (Value s)
```

which can be specialized to `Text`, but selecting `string_agg(foo, bar)` from an empty record set `foo` results in `NULL`. I noticed the other aggregations return `Maybe`, forcing the user to deal with this explicitly. Should `stringAgg` do the same?

Philonous:
This is (some of) the Postgres-specific parts from #75. I have implemented a more generic aggregation function [unsafeSqlAggregateFunction](https://github.com/bitemyapp/esqueleto/compare/master...Philonous:postgresql-additions?expand=1#diff-c901387a600e739ea0b320c45ad0882bR59) that allows setting `ORDER BY` clauses and made `arrayAgg` and `stringAgg` special cases.

This also fixes #68 and a related bug in `arrayAgg`. Note that the types of `arrayAgg` and `stringAgg` had to be changed because of that. (An alternative fix would be to add `COALESCE` clauses to those functions, which would make them more convenient to use and avoid changing the types, but it would break with the mantra that esqueleto translates straightforwardly into SQL)

[Tests are included.](https://github.com/bitemyapp/esqueleto/compare/master...Philonous:postgresql-additions?expand=1#diff-da86d9f90059b54422c926fa2d992dd3R245)

### DIFF
--- a/src/Database/Esqueleto/Internal/Sql.hs
+++ b/src/Database/Esqueleto/Internal/Sql.hs
@@ -17,7 +17,7 @@
 module Database.Esqueleto.Internal.Sql
   ( -- * The pretty face
     SqlQuery
-  , SqlExpr
+  , SqlExpr(..)
   , SqlEntity
   , select
   , selectSource
@@ -35,17 +35,25 @@ module Database.Esqueleto.Internal.Sql
   , unsafeSqlFunction
   , unsafeSqlExtractSubField
   , UnsafeSqlFunctionArgument
+  , OrderByClause
   , rawSelectSource
   , runSource
   , rawEsqueleto
   , toRawSql
   , Mode(..)
+  , NeedParens(..)
   , IdentState
   , initialIdentState
   , IdentInfo
   , SqlSelect(..)
   , veryUnsafeCoerceSqlExprValue
   , veryUnsafeCoerceSqlExprValueList
+  -- * Helper functions
+  , makeOrderByNoNewline
+  , uncommas'
+  , parens
+  , toArgList
+  , builderToText
   ) where
 
 import Control.Arrow ((***), first)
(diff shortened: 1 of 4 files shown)


src/Database/Esqueleto/Internal/Sql.hs
```diff
@@ -17,7 +17,7 @@
 module Database.Esqueleto.Internal.Sql
   ( -- * The pretty face
     SqlQuery
-  , SqlExpr
+  , SqlExpr(..)
```
bitemyapp: Is it necessary to export `SqlExpr`'s constructors?

test/PostgreSQL/Test.hs
```diff
@@ -330,3 +554,11 @@ run_worker act = withConn $ runSqlConn (migrateIt >> act)
 withConn :: RunDbMonad m => (SqlBackend -> R.ResourceT m a) -> m a
 withConn =
   R.runResourceT . withPostgresqlConn "host=localhost port=5432 user=esqutest password=esqutest dbname=esqutest"
+
+-- | Show the SQL generated by a query
+showQuery :: (Monad m, ES.SqlSelect a r, BackendCompatible SqlBackend backend)
```
bitemyapp: This is pretty handy, I've wanted a glued-together `rawSql/toRawSql` like this before. Hopefully people are smart and don't try to use this for anything other than logging/debugging.

Philonous: This is only in the test module and not exported.

src/Database/Esqueleto/Internal/Sql.hs
Philonous: Unfortunately it was the only way I could figure out to make [unsafeSqlAggregateFunction](https://github.com/bitemyapp/esqueleto/pull/76/files#diff-c901387a600e739ea0b320c45ad0882bR67) work. 

Besides, is there a specific reason not to export it from an .Internal module? The usual recommendation is to expose as much as possible through them as it allows for cleaner implementations of custom functions like this  and .Internal modules usually come with an implied "here be dragons, user discretion is advised" anyway.

bitemyapp: @Philonous that's fine by me, I just wanted to make sure I understood.

paf31:
The exported `SqlExpr` constructor turned out to be quite handy for me, for what it's worth. It allowed me to write some coercions I needed which would have been impossible with the exported API.

bitemyapp:
@paf31 makes sense, thanks for reporting back! It helps to know stuff like this.

PR #83, opened 2018-03-24
got it to build

parsonsmatt:
I convinced myself pretty quickly that should-not-typecheck was fucky, so I moved the test into the example.

The example builds with stack now and has it's own stack.yaml. If you uncomment the "doesn'twork" thing then you get a type error.

SqlWriteT can do reads. SqlReadT can't do writes.


examples/Main.hs
```diff
@@ -86,6 +87,12 @@ getAdults =
     where_ (p ^. PersonAge >=. just (val 18))
     return p
 
+-- shouldn'twork
+--   :: (MonadIO m, MonadLogger m)
+--   => SqlReadT m ()
+-- shouldn'twork = delete $ from $ \p ->
+--   where_ (p ^. PersonAge >=. just (val 18))
```
parsonsmatt: uncomment for glorious failure

src/Database/Esqueleto/Internal/Sql.hs
```diff
@@ -972,77 +972,26 @@ update
...
+  :: ( BackendCompatible SqlBackend (PersistEntityBackend record)
      , BackendCompatible SqlBackend backend
-     , BackendCompatible SqlBackend (BaseBackend backend)
-     , SqlBackendCanWrite backend
-     -- , BackendCompatible SqlBackend (PersistEntityBackend val)
-     -- , PersistQueryWrite backend
-     -- , PersistUniqueWrite backend
-     -- , PersistStoreWrite backend
+     , PersistEntity record
+     , MonadIO m
```
parsonsmatt: these are the constraints that got it compiling nicely

PR #84, opened 2018-04-02
Fix build under GHC 8.4

k-bx:
Tests don't build because of https://github.com/lpsmith/postgresql-simple/issues/252


bitemyapp:
Sweet, thank you @k-bx! Nice seeing you around too :)

PR #94, opened 2018-06-22
Define applicative instance for Value

danbroooks:


Cmdv:
I've been requiring this as well

v0d1ch:
I am not sure why Applicative instance was not there in the first place ?  Is there any reason not to have it ? Personally I think it is a handy thing to have lying around. 🚢 it

PR #99, opened 2018-08-01
Fix "wike" and "envoke" typos in README.md

mattaudesse:
Opinions seem to differ regarding whether "envoke" is incorrect or just
unusual in modern English (but please educate me if neither is the case):

https://correct-spelling.com/envoke-or-invoke/
https://en.wiktionary.org/wiki/envoke

Thanks for `Esqueleto`!

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,7 +262,7 @@ brew install postgresql
 brew install libpq
 ```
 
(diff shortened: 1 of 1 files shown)


bitemyapp:
Tell @meteficha - I just keep the lights on :)

Thank you!

meteficha:
Thanks for keeping it in good shape, @bitemyapp!

PR #111, opened 2019-01-03
Add between

### LINKED ISSUE #34: Add: BETWEEN
What is your opinion about adding BETWEEN ... AND ...?
I find this feature useful while writing SQL, but esqueleto lacks it.

ibarrae:
Fixes #34 

Please, let me know if there's anything else to be done/fixed regarding this issue 👍

### DIFF
--- a/src/Database/Esqueleto/Internal/Language.hs
+++ b/src/Database/Esqueleto/Internal/Language.hs
@@ -354,7 +354,7 @@ class (Functor query, Applicative query, Monad query) =>
 
   -- | @BETWEEN@ operator
   --
-  -- /Since: 2.6.0/
+  -- /Since: 2.8.0/
   between :: PersistField typ => expr (Value typ) -> (expr (Value typ), expr (Value typ)) -> expr (Value Bool)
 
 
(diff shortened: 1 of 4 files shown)


src/Database/Esqueleto/Internal/Language.hs
```diff
@@ -352,6 +352,11 @@ class (Functor query, Applicative query, Monad query) =>
   (/.)  :: PersistField a => expr (Value a) -> expr (Value a) -> expr (Value a)
   (*.)  :: PersistField a => expr (Value a) -> expr (Value a) -> expr (Value a)
 
+  -- | @BETWEEN@ operator
+  --
+  -- /Since: 2.6.0/
```
parsonsmatt:
```suggestion
  -- /Since: 2.8.0/
```

This will be released as part of 2.8.0, not 2.6.0

parsonsmatt:
Unfortunately this implementation doesn't work with non-`val` values. Here's a failing test case:

```
    it "works for a SqlExpr value" $ run $ do
        p1e  <- insert' p1
        _    <- insert' p2
        _    <- insert' p3
        ret <-
          select $
          from $ \p -> do
          where_ $
            just (p ^. PersonFavNum)
              `between`
                (p ^. PersonAge, p ^.  PersonWeight)
        liftIO $ ret `shouldBe` []
```

This fails with a format error. It should eventually succeed and then the `[]` return value should be filled in appropriately.

I can't figure out how to get this PR unmerged, so you will have to open a new PR :\

PR #113, opened 2019-02-01
Remove the type class

parsonsmatt:
This PR removes the type class from `esqueleto` and simplifies all relevant types.


FintanH:
Fantastic :heart:

parsonsmatt:
@bitemyapp This is a big breaking change. Think we should lodge this in a 3.0 release or 2.8?

bitemyapp:
@parsonsmatt Based on http://hackage.haskell.org/package/esqueleto, I'd say this is a 3.0.

parsonsmatt:
@AKurilin I know y'all use this library a lot. Would this change harm your code base? I'm going to check with Lumi's codebase as well.

For some additional motivation for this change, I'm working on some complicated SQL logic. Somehow, the type is getting fixed to `Esqueleto Ratio SqlQuery SqlBackend`, and it's giving *all kinds* of totally useless errors. I know this is a simple type error. If the type were concrete, then I'd get an immediate pointer to where `SqlExpr` is getting confused with `Ratio`.

eborden:
@parsonsmatt This is a breaking change for us (Freckle). Though it is a breaking change I'd welcome.

cdparks:
I think there are only two places we'd need to update. All of our other helpers use concrete `SqlQuery` and `SqlExpr`.

parsonsmatt:
@eborden @cdparks Is it just a mechanical translation from `Esqueleto e q b => e x -> q (e y)` to `SqlExpr x -> SqlQuery (SqlExpr y)`, or is there some functionality that would be lost with this?

cdparks:
Yeah, that's exactly the change we'd have to make. We wouldn't lose any functionality, and the upgrade path is like a one-liner.

andrewthad:
I use a little bit of esqueleto in about a dozen different projects. I've never seen any value in having the typeclass since there's only one possible instance of it. I would be happy to see it go away.

eborden:
@parsonsmatt FYI, @cdparks has already fixed this in our codebase. All uses were unnecessary. The transition will now be painless.

parsonsmatt:
Thanks @eborden @cdparks :)

Issue #110, opened 2018-12-29
Building haddock for ghc-8.4+ is broken


brandon-leapyear:
```
$ cat .stack-work/logs/esqueleto-2.6.0.log
Configuring esqueleto-2.6.0...
Preprocessing library for esqueleto-2.6.0..
Building library for esqueleto-2.6.0..
Preprocessing library for esqueleto-2.6.0..
Running Haddock on library for esqueleto-2.6.0..
Haddock coverage:
haddock: internal error: internal: extractDecl
CallStack (from HasCallStack):
  error, called at utils/haddock/haddock-api/src/Haddock/Interface/Create.hs:1116:12 in main:Haddock.Interface.Create
```

This is broken both in `lts-12.24` and `lts-13.0`. It's probably a haddock bug, but I can't figure out why it's happening.

I suspect it has something to do with selectively exporting functions from [`Esqueleto`](https://github.com/bitemyapp/esqueleto/blob/master/src/Database/Esqueleto.hs#L41), but I don't know for sure.

brandon-leapyear:
I added debug code and got:
```
haddock: panic! (the 'impossible' happened)
  (GHC version 8.4.4 for x86_64-apple-darwin):
	extractDecl

Unhandled decl for SqlBackendKey:
    data family BackendKey backend :: *
Call stack:
    CallStack (from HasCallStack):
      callStackDoc, called at compiler/utils/Outputable.hs:1150:37 in ghc:Outputable
      pprPanic, called at src/Haddock/Interface/Create.hs:1100:12 in haddock-api-2.20.0-9FnywoxlUeu2IGqOI4hPVv:Haddock.Interface.Create
```
So this could be a bug in haddock/persistent

bitemyapp:
I don't think this is in our wheelhouse. I'm closing for now, re-open if I misunderstood or you learn something new.

parsonsmatt:
unfortunately I believe this will block putting `esqueleto` back on Stackage. I'm reopening

FintanH:
Is this causing the docs to not appear on Hackage? http://hackage.haskell.org/package/esqueleto-2.6.1/docs/Database-Esqueleto.html

![image](https://user-images.githubusercontent.com/3503528/53802936-16d21d80-3f3b-11e9-935a-c342ce767061.png)

parsonsmatt:
hmmm, I thought that this was going to be fixed by #114.

parsonsmatt:
[...]

PR #114, opened 2019-02-17
Export Persistent.Sql module explicitly

ChrisCoffey:
I ran afoul of https://github.com/bitemyapp/esqueleto/issues/110, but the advice in https://github.com/haskell/haddock/issues/979#issuecomment-459946930 suggesting an explicit export list rather than module export addresses the problem.

Its a pretty ugly solution & should be undone once [the ghc bug](https://gitlab.haskell.org/ghc/ghc/merge_requests/276) ships. But, it lets us build the docs...


ChrisCoffey:
Yes, that is the entire export list as specified in the `.hi` file.

parsonsmatt:
Wonderful, thanks! :joy:

PR #119, opened 2019-03-11
add postgress full text search operator

andrewthad:
work in progress

### DIFF
--- a/src/Database/Esqueleto/PostgreSQL.hs
+++ b/src/Database/Esqueleto/PostgreSQL.hs
@@ -15,6 +15,7 @@ module Database.Esqueleto.PostgreSQL
   , stringAgg
   , stringAggWith
   , maybeArray
+  , (@@.)
   , chr
   , now_
   , random_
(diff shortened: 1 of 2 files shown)


src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -143,6 +144,14 @@ stringAgg ::
   -> SqlExpr (Value (Maybe s)) -- ^ Concatenation.
 stringAgg expr delim = stringAggWith AggModeAll expr delim []
```
parsonsmatt:
```suggestion
-- | Full text search operator.
--
-- @since 2.8.0
```

andrewthad:
Although this currently works correctly, it’s probably better to have the second argument to this function accept something different.  Maybe an algebraic data types for constructing  various full text search predicates.

parsonsmatt:
Good point. Looks like it supports `&`, `|`, and `!`. It should be relatively straightforward to write that.

We may also want to incorporate more options for the `tsvector` type, too: https://www.postgresql.org/docs/9.5/datatype-textsearch.html 

It may be reasonable to have `@@.` for the simple case and `+@@.` (where `+` represents "extra") that's typed like `(+@@.) :: SqlExpr (Value TsVector) -> SqlExpr (Value TsQuery) -> SqlExpr Bool`.

bitemyapp:
Gonna close this for now, I'd be happy to merge a complete version of this if you pick this back up and take it home.

PR #120, opened 2019-04-06
Update README.md

StevenXL:
The application of `just` is unnecessary, and actually leads to a type error. The `?.` does all the work. ;)


bitemyapp:
Thanks!

PR #122, opened 2019-04-18
persistent-2.10 support

parsonsmatt:
Unfortunately a breaking change due to reexport of `PersistentImport`. That should probably be different.


parsonsmatt:
So, we can make this *not* a breaking change by CPPing in the export depending on the persistent version. 

```haskell
#if !MIN_VERSION_persistent(2,10,0)
type Connection = SqlBackend
{-# DEPRECATED Connection "stawp it" #-}
#endif
```

Basically, redefining the type synonyms that were deleted and giving new deprecation warnings.

I don't like this idea, fwiw.

parsonsmatt:
@bitemyapp Want to release this now, or would you like to collect some more stuff for the breaking change release?

parsonsmatt:
this gonna be part of 3.0

PR #126, opened 2019-06-17
sub_select produces invalid SQL sometimes

parsonsmatt:
This PR currently adds a failing test to demonstrate the problem. Will open an issue later.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@
 .cabal-sandbox/
 cabal.sandbox.config
 .hspec-failures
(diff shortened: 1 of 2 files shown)


bitemyapp:
>`SqlError {sqlState = "42883", sqlExecStatus = FatalError, sqlErrorMsg = "function array_remove(record) does not exist", sqlErrorDetail = "", sqlErrorHint = "No function matches the given name and argument types. You might need to add explicit type casts."}`

PR #127, opened 2019-06-26
Add between

ibarrae:
Between was added by myself on #111 but it was reverted on #123 due to the fact that new `between` clause was not working with non-`val` values as @parsonsmatt mentioned [here](https://github.com/bitemyapp/esqueleto/pull/111#issuecomment-485042071).

I changed that and also included some tests to cover that case.

Please, let me know if there's another missing case so I can also address it :smiley:

### DIFF
--- a/src/Database/Esqueleto/Internal/Internal.hs
+++ b/src/Database/Esqueleto/Internal/Internal.hs
@@ -459,12 +459,19 @@ not_ (ECompositeKey _) = throw (CompositeKeyErr NotError)
 --
 -- /Since: 3.0.0/
 between :: PersistField a => SqlExpr (Value a) -> (SqlExpr (Value a), SqlExpr (Value a)) -> SqlExpr (Value Bool)
-a `between` (ERaw p1 f, ERaw p2 g) = unsafeSqlBinOp " BETWEEN " a $ ERaw Never $ \x ->
-  let (v1, fv) = f x
-      (v2, gv) = g x
-  in  (parensM p1 v1 <> " AND " <> parensM p2 v2, fv ++ gv)
-_ `between` _                      = throw $ CompositeKeyErr BetweenError
-
+a@ERaw{} `between` (ERaw p1 f, ERaw p2 g) =
+  unsafeSqlBinOp " BETWEEN " a $ ERaw Never $ \x ->
+    let (v1, fv) = f x
+        (v2, gv) = g x
+    in  (parensM p1 v1 <> " AND " <> parensM p2 v2, fv ++ gv)
+a `between` (b, c)                        =
+  let construct :: PersistField a => SqlExpr (Value a) -> SqlExpr (Value a)
+      construct (ERaw p f) = throw $ CompositeKeyErr BetweenError
+      construct (ECompositeKey f) = ERaw Parens $ \i -> (uncommas $ f i, mempty)
+      valA = construct a
+      valB = construct b
+      valC = construct c
+   in valA >=. valB &&. valA <=. valC
 
 random_  :: (PersistField a, Num a) => SqlExpr (Value a)
 random_  = unsafeSqlValue "RANDOM()"
--- a/src/Database/Esqueleto/Internal/Internal.hs
+++ b/src/Database/Esqueleto/Internal/Internal.hs
@@ -455,6 +455,16 @@ not_ (ECompositeKey _) = throw (CompositeKeyErr NotError)
 (*.)  :: PersistField a => SqlExpr (Value a) -> SqlExpr (Value a) -> SqlExpr (Value a)
 (*.)  = unsafeSqlBinOp " * "
 
+-- | @BETWEEN@.
+--
+-- /Since: 3.0.0/
+between :: PersistField a => SqlExpr (Value a) -> (SqlExpr (Value a), SqlExpr (Value a)) -> SqlExpr (Value Bool)
+a `between` (ERaw p1 f, ERaw p2 g) = unsafeSqlBinOp " BETWEEN " a $ ERaw Never $ \x ->
+  let (v1, fv) = f x
+      (v2, gv) = g x
+  in  (parensM p1 v1 <> " AND " <> parensM p2 v2, fv ++ gv)
+_ `between` _                      = throw $ CompositeKeyErr BetweenError
+
 
 random_  :: (PersistField a, Num a) => SqlExpr (Value a)
 random_  = unsafeSqlValue "RANDOM()"
@@ -1249,6 +1259,7 @@ data CompositeKeyError =
   | MakeSetError
   | MakeWhereError
   | MakeHavingError
+  | BetweenError
   deriving (Show)
 
 data UnexpectedCaseError =
@@ -2785,4 +2796,3 @@ insertSelect = void . insertSelectCount
 insertSelectCount :: (MonadIO m, PersistEntity a) =>
   SqlQuery (SqlExpr (Insertion a)) -> SqlWriteT m Int64
 insertSelectCount = rawEsqueleto INSERT_INTO . fmap EInsertFinal
-
--- a/src/Database/Esqueleto/Internal/Internal.hs
+++ b/src/Database/Esqueleto/Internal/Internal.hs
@@ -466,7 +466,7 @@ a@ERaw{} `between` (ERaw p1 f, ERaw p2 g) =
     in  (parensM p1 v1 <> " AND " <> parensM p2 v2, fv ++ gv)
 a `between` (b, c)                        =
   let construct :: PersistField a => SqlExpr (Value a) -> SqlExpr (Value a)
-      construct (ERaw p f) = throw $ CompositeKeyErr BetweenError
+      construct ERaw{}            = throw $ CompositeKeyErr BetweenError
       construct (ECompositeKey f) = ERaw Parens $ \i -> (uncommas $ f i, mempty)
       valA = construct a
       valB = construct b
--- a/src/Database/Esqueleto/Internal/Internal.hs
+++ b/src/Database/Esqueleto/Internal/Internal.hs
@@ -1647,9 +1647,12 @@ unsafeSqlBinOp op (ERaw p1 f1) (ERaw p2 f2) = ERaw Parens f
                 , vals1 <> vals2 )
 unsafeSqlBinOp op a b = unsafeSqlBinOp op (construct a) (construct b)
     where construct :: SqlExpr (Value a) -> SqlExpr (Value a)
-          construct (ERaw p f)        = ERaw Never $ \info ->
+          construct (ERaw p f)        = ERaw Parens $ \info ->
             let (b1, vals) = f info
-             in (parensM p b1, vals)
+                build ("?", [PersistList vals']) =
+                  (uncommas $ replicate (length vals') "?", vals')
+                build expr = expr
+             in  build (parensM p b1, vals)
           construct (ECompositeKey f) =
             ERaw Parens $ \info -> (uncommas $ f info, mempty)
 {-# INLINE unsafeSqlBinOp #-}
--- a/src/Database/Esqueleto/Internal/Internal.hs
+++ b/src/Database/Esqueleto/Internal/Internal.hs
@@ -459,19 +459,7 @@ not_ (ECompositeKey _) = throw (CompositeKeyErr NotError)
 --
 -- /Since: 3.0.0/
 between :: PersistField a => SqlExpr (Value a) -> (SqlExpr (Value a), SqlExpr (Value a)) -> SqlExpr (Value Bool)
-a@ERaw{} `between` (ERaw p1 f, ERaw p2 g) =
-  unsafeSqlBinOp " BETWEEN " a $ ERaw Never $ \x ->
-    let (v1, fv) = f x
-        (v2, gv) = g x
-    in  (parensM p1 v1 <> " AND " <> parensM p2 v2, fv ++ gv)
-a `between` (b, c)                        =
-  let construct :: PersistField a => SqlExpr (Value a) -> SqlExpr (Value a)
-      construct ERaw{}            = throw $ CompositeKeyErr BetweenError
-      construct (ECompositeKey f) = ERaw Parens $ \i -> (uncommas $ f i, mempty)
-      valA = construct a
-      valB = construct b
-      valC = construct c
-   in valA >=. valB &&. valA <=. valC
+a `between` (b, c) = a >=. b &&. a <=. c
 
 random_  :: (PersistField a, Num a) => SqlExpr (Value a)
 random_  = unsafeSqlValue "RANDOM()"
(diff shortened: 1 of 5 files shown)


src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -461,6 +461,16 @@ not_ (ECompositeKey _) = throw (CompositeKeyErr NotError)
...
 
+-- | @BETWEEN@.
+--
+-- /Since: 3.0.0/
+between :: PersistField a => SqlExpr (Value a) -> (SqlExpr (Value a), SqlExpr (Value a)) -> SqlExpr (Value Bool)
+a `between` (ERaw p1 f, ERaw p2 g) = unsafeSqlBinOp " BETWEEN " a $ ERaw Never $ \x ->
+  let (v1, fv) = f x
+      (v2, gv) = g x
+  in  (parensM p1 v1 <> " AND " <> parensM p2 v2, fv ++ gv)
+_ `between` _                      = throw $ CompositeKeyErr BetweenError
```
parsonsmatt: I think that we can properly handle this instead of throwing an error. `between` in Postgresql handles composite values just fine:

```sql
[local] matt@matt=# select ((5, 3) between (3, 2) and (6, 4));
 ?column? 
----------
 t
(1 row)
```

as does sqlite:

```sql
sqlite> select ((5, 3) between (3, 2) and (6, 4));
1
```

However, it doesn't work for MySQL -_-

```sql
mysql> select ((5, 3) between (3, 2) and (6, 4));
ERROR 1241 (21000): Operand should contain 1 column(s)
```

Sigh. This is the trouble with being generic over various SQL backends.

Especially annoying is the fact that MySQL works with the "dumb" implementation:

```sql
mysql> select ((5, 3) > (3, 2) and (5, 3) < (6, 4));
+---------------------------------------+
| ((5, 3) > (3, 2) and (5, 3) < (6, 4)) |
+---------------------------------------+
|                                     1 |
+---------------------------------------+
1 row in set (0.00 sec)
```

So, we have a few options:

1. Define `between` that uses the SQL keyword `between` with composite keys and reexport fromt he Postgresql and Sqlite modules. Define `between` that uses the expanded form for MySQL and export from the Mysql module.
2. Define `between` that uses `between` for scalar values and the expanded form for composite values. We'd need to ensure that it has the same semantics on the edges/boundaries.

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1255,6 +1265,7 @@ data CompositeKeyError =
   | MakeSetError
   | MakeWhereError
   | MakeHavingError
+  | BetweenError
```
parsonsmatt: I think we can remove this when `CompositeKey` is supported.

test/Common/Test.hs
```diff
@@ -630,6 +630,28 @@ testSelectWhere run = do
...
+      run $ do
+        _ <- insert' p1 >> insert' p2 >> insert' p3
+        ret <-
+          select $
+          from $ \p -> do
+          where_ $
+            just (p ^. PersonFavNum)
+              `between`
+                (p ^. PersonAge, p ^.  PersonWeight)
+        liftIO $ ret `shouldBe` []
```
parsonsmatt: Let's write some tests for composite keys, too. There's no reason we can't support them.

src/Database/Esqueleto/Internal/Internal.hs
ibarrae: I didn't know about this, thanks @parsonsmatt will look into this also :smile:

src/Database/Esqueleto/Internal/Internal.hs
ibarrae: Shouldn't we throw this error for MySQL since it can't be used anyways as you just mentioned?

src/Database/Esqueleto/Internal/Internal.hs
parsonsmatt: It could also be simply defined as:

```haskell
between a (x, y) = a >. x &&. a <. y
```

which should work for all backends. Not sure if that exactly matches the semantics of `BETWEEN` though.

src/Database/Esqueleto/Internal/Internal.hs
parsonsmatt: I'd rather make it work for MySQL if at all possible, even if we have to do the expanded form.

src/Database/Esqueleto/Internal/Internal.hs
ibarrae: Maybe if it makes a mixture of option 2 and your last comment?
How about to make it work as it is now for scalar values and use `a >. x && a <.y` for composite keys?

parsonsmatt: That works for me!

We need to ensure that we use the right operators - I'm not exactly sure what happens if you do `(1, 2) BETWEEN (1, 2) AND (2,3)` or `(2, 3) BETWEEN (1, 2) AND (2, 3)`.

ibarrae: I will test this and get back to you!

ibarrae:
Hey, @parsonsmatt I just pushed some changes but it maintains `BetweenError` because I figured out while doing some tests that when you mix scalar and composite values it also fails on different databases (using `between` and `>=` `<=` accordingly ) and would lead to runtime errors, so I think it's better to throw an error if this scenario occurs.. Summing up, it works when all values for between are scalar or composite only, but it cannot be scalar and composite. Let me know what you think of this.

```sql
mysql> select * from point where ((x, y) >= (1) and (x,y) <= (3,4));
ERROR 1241 (21000): Operand should contain 2 column(s)
```

```sql
postgres> select 1 from table where (1, 2) >= 1 and (1, 2) <= (2,3);
select 1 from table where (1, 2) between 1 and (3, 4); 
ERROR:  operator does not exist: record >= integer
```

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -461,6 +461,23 @@ not_ (ECompositeKey _) = throw (CompositeKeyErr NotError)
...
+-- /Since: 3.0.0/
+between :: PersistField a => SqlExpr (Value a) -> (SqlExpr (Value a), SqlExpr (Value a)) -> SqlExpr (Value Bool)
+a@ERaw{} `between` (ERaw p1 f, ERaw p2 g) =
+  unsafeSqlBinOp " BETWEEN " a $ ERaw Never $ \x ->
+    let (v1, fv) = f x
+        (v2, gv) = g x
+    in  (parensM p1 v1 <> " AND " <> parensM p2 v2, fv ++ gv)
+a `between` (b, c)                        =
+  let construct :: PersistField a => SqlExpr (Value a) -> SqlExpr (Value a)
+      construct (ERaw p f) = throw $ CompositeKeyErr BetweenError
```
parsonsmatt: I don't think this is necessary either -

`ERaw` happens when we do `unsafeSqlFunction`. However, we have a guarantee from the types that all the expressions are `SqlExpr (Value a)`. This means that, as far as Haskell is concerned, we have the same type for everything.

If it is a `ECompositeKey` on one side, then it's going to either be `ECompositeKey` on the other sides, *or* it is going to be a `SqlExpr` that represents the same type of thing.

Like we might call this with: `SqlExpr (Value Rational)` or `SqlExpr (Value UserCompositeKey)`. But if we do, then we know that we'll get it right.

So, I think for this branch, we can just write:

```
a `between` (b, c) = a >=. b &&. a <=. c
```

Does this pass all tests?

ibarrae: If we use that directly it fails because `>=.`, `<=.` and `&&.`  depend on `unsafeSqlBinOp` and it [doesn't support composite keys](https://github.com/bitemyapp/esqueleto/blob/a0274e377b7f0eaa4eaf1d2f254e2d803be9ed7e/src/Database/Esqueleto/Internal/Internal.hs#L1650) :disappointed_relieved:

ibarrae: Also, while I was doing tests and while writing queries I normally use `val` function but that just passes every value for the placeholder `?` even composite keys making them `ERaw` values, making errors at runtime.

For example this will also fail:
```haskell
select $
  from $ \p -> do
    where_ $
      p ^. PointId
        `between`
           ( val $ PointKey 1 2
           , val $ PointKey 5 6)
```

parsonsmatt: Right, `val` does `ERaw` - you'd get an `ECompositKey` from projecting from a point, like `p ^. PointId`

I think fixing the `unsafeSqlBinOp` to support composite keys makes sense to do here.

ibarrae: That sounds good to me :smile: but when we have the scenario that I described before, meaning that for `unsafeSqlBinOp` to handle `ERaw` and `ECompsiteKey` or backward, should it keep failing or how it should be handled?

parsonsmatt: There's pretty much no reason to fail that - we should just render the two arguments on either side. The function is already labeled as `unsafe` so failing at runtime with a SQL syntax error isn't outside the realm of possibility.

ibarrae: Great! I'll work on that then, thanks for the help :smile:

ibarrae:
I updated `unsafeSqlBinOp`, would you mind taking a look again @parsonsmatt

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1647,7 +1651,13 @@ unsafeSqlBinOp op (ERaw p1 f1) (ERaw p2 f2) = ERaw Parens f
...
              in ( parensM p1 b1 <> op <> parensM p2 b2
                 , vals1 <> vals2 )
-unsafeSqlBinOp _ _ _ = throw (CompositeKeyErr SqlBinOpError)
+unsafeSqlBinOp op a b = unsafeSqlBinOp op (construct a) (construct b)
+    where construct :: SqlExpr (Value a) -> SqlExpr (Value a)
+          construct (ERaw p f)        = ERaw Never $ \info ->
+            let (b1, vals) = f info
+             in (parensM p b1, vals)
+          construct (ECompositeKey f) =
+            ERaw Parens $ \info -> (uncommas $ f info, mempty)
```
parsonsmatt: :tada: 

This looks good to me! Can we write a test for it?

I think something like:

```haskell
select $
from $ \p -> do
where_ (p ^. SomeCompositeField `between` (val $ MkCompositeField 1 2, val $ MkCompositeField 3 4)
```

would do it.

ibarrae: I added a test for it [here](https://github.com/bitemyapp/esqueleto/blob/34d3805ce0d0c15b834975cebde898eeff64a65e/test/Common/Test.hs#L655). Is that enough?

parsonsmatt: You should be able to do `val $ PointKey 1 2` instead of manually calling the `ECompositeKey` constructor

ibarrae: That would lead to failing a failing test as we discussed before, due to the fact that `val` wraps everything on `ERaw` ADT :disappointed:

parsonsmatt: I thought your change to `unsafeSqlBinOp` fixed that?

ibarrae: Not really, with this change `unsafeSqlBinOp` can know handle composite keys but you still have to manually specify it as composite as on the test, with `val` at its current state is impossible to create `ECompositeKey` from any `persistent` composite primary key.

parsonsmatt: :thinking: 

Users should be able to do this seamlessly - comparing a projected composite key and a `val` composite key should work. So I'm curious what happens now when you run your code with the `val` in a `between` function

ibarrae: When using scalar values only it works, for only composite values works two, the problem here is when you use a mixture of both and when using `val` it transforms everything on scalar values or `ERaw`, that's more of an issue of `val` instead of `unsafeSqlBinOp` in my opinion, and as you mentioned [here](https://github.com/bitemyapp/esqueleto/pull/127#discussion_r298259357) at least is my understanding that users should be aware of this, that's why `unsafeSqlBinOp` will not `throw` any error but will fail at runtime with SQL syntax error

ibarrae: Hey @parsonsmatt, have you been able to review the last comment or maybe came up with another idea for these changes?

parsonsmatt: Thank you for pinging me again! I need to think about this for a bit and I've put a note on my todo list to check on it.

Also pinging @bitemyapp and maybe @FintanH to provide input if they have time/energy to do so

parsonsmatt: I've got this cloned locally and I'm going to try writing some tests for expected behavior with `val $ CompositeKey 'a' 'b'` etc to see if I can finagle it into a working place

ibarrae:
Hey @parsonsmatt I think I got this correctly this time, could you take a look when possible, please and also, CI seems to be failing due to a time out, could you restart it if possible too?

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -461,6 +461,11 @@ not_ (ECompositeKey _) = throw (CompositeKeyErr NotError)
 (*.)  :: PersistField a => SqlExpr (Value a) -> SqlExpr (Value a) -> SqlExpr (Value a)
 (*.)  = unsafeSqlBinOp " * "
 
+-- | @BETWEEN@.
+--
+-- /Since: 3.0.0/
```
parsonsmatt: This isn't going to be right, since `3.0.0` is already released. It'll batch up with 3.1.0.

```suggestion
-- @since 3.1.0
```

test/Common/Test.hs
```diff
@@ -630,6 +630,54 @@ testSelectWhere run = do
...
+            insert_ $ Point 1 2 ""
+            ret <-
+              select $
+              from $ \p -> do
+              where_ $
+                p ^. PointId
+                  `between`
+                    ( EI.ECompositeKey $ const ["3", "4"]
+                    , EI.ECompositeKey $ const ["5", "6"] )
+            liftIO $ ret `shouldBe` []
```
parsonsmatt: Wonderful tests, thanks! :smile:

test/Common/Test.hs
```diff
@@ -630,6 +630,54 @@ testSelectWhere run = do
...
+          run $ do
+            insert_ $ Point 1 2 ""
+            ret <-
+              select $
+              from $ \p -> do
+              where_ $
+                p ^. PointId
+                  `between`
+                    ( val $ PointKey 1 2
+                    , val $ PointKey 5 6 )
```
parsonsmatt: Awesome!!! This is great.

ibarrae:
@parsonsmatt Done! I updated the version number and updated the changelog entry, thanks a lot for your help in this :smile:

parsonsmatt:
Thank *you*! You've written a new feature, discovered a bug, and fixed that bug all in one go :smile:

PR #128, opened 2019-07-29
PostgreSQL JSON Operators

Vlix:
Added all JSON operators in the `Database.Esqueleto.PostgreSQL.JSON` module and wrote documentation and tests for all operators.

Also increased version to 3.1.0


Vlix:
Ugh... I can't get Travis to use PSQL 10 > .<

Any idea what's going on?

src/Database/Esqueleto/PostgreSQL/JSON.hs
```diff
@@ -0,0 +1,590 @@
...
+      in case it doesn't type check.
+      * As long as the previous operator's resulting value is
+      'Maybe' 'Data.Aeson.Value', any other JSON operator can be
+      used to transform the JSON further.
+      (e.g. @'[1,2,3]' -> 1 \@> 2@)
+
+  /The PostgreSQL version the functions work with are included/
+  /in their description./
+
+  /Since: 3.1.0/
```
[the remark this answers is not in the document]
Vlix: I've used the `/Since: 3.1.0/` notation, since that's what's used throughout the entire package, but I can easily fix it, if you'd prefer the `@` notation?

src/Database/Esqueleto/PostgreSQL/JSON.hs
```diff
@@ -0,0 +1,590 @@
...
+      'Maybe'. This also makes it easier to chain them.
+      Just use the 'just' function to lift any non-'Maybe' JSON values
+      in case it doesn't type check.
+      * As long as the previous operator's resulting value is
+      'Maybe' 'Data.Aeson.Value', any other JSON operator can be
+      used to transform the JSON further.
+      (e.g. @'[1,2,3]' -> 1 \@> 2@)
+
+  /The PostgreSQL version the functions work with are included/
+  /in their description./
```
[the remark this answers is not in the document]
Vlix: 😳 You make me blush

parsonsmatt:
Hmm, the [docs](https://docs.travis-ci.com/user/database-setup/#postgresql) are a little ambiguous, but I think you should remove the `postgresql` in `services` and just use the one in `addons`

src/Database/Esqueleto/PostgreSQL/JSON.hs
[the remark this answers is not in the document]
parsonsmatt: Yeah I think the `@since` notation is newer than the package. It potentially has some semantic meaning to Haddock so it's probably better to add it.

parsonsmatt:
This is the example for using postgresql-10:

```
addons:
  postgresql: "10"
  apt:
    packages:
    - postgresql-10
    - postgresql-client-10
```

What you've got now is:

```
addons:
  postgresql: "10"
  apt:
    packages:
      - libgmp-dev
      - postgresql-client-10
      - postgresql-server-dev-all
```

Only thing I can see is `postgresql-server-dev-all` vs `-postgresql-10`

Vlix:
Yeah, I'm not sure what the `libqmp-dev` or `postgresql-server-dev-all` is for... but I guess I'll keep them in?

Vlix:
Seems PostgreSQ 10 runs on port 5433 by default?

bitemyapp:
@Vlix this is looking great, I love the documentation!

Vlix:
> @Vlix this is looking great, I love the documentation!

Thanks! Not just for the compliment, but also for `Haskell Programming: From First Principles`. It's what really got me going a few years back :)

Vlix:
I think I finally fixed the dependency stuff (same problem as with `persistent` when it went to 2.10.0)

parsonsmatt:
awesome, thanks, I'll review today

src/Database/Esqueleto/PostgreSQL/JSON.hs
```diff
@@ -0,0 +1,618 @@
...
+-- @
+--     | Type | Description                                |  Example                                         | Example Result
+-- ----+------+--------------------------------------------+--------------------------------------------------+----------------
+--  -> | int  | Get JSON array element (indexed from zero) | '[{"a":"foo"},{"b":"bar"},{"c":"baz"}]'::json->2 | {"c":"baz"}
+--  -> | text | Get JSON object field by key               | '{"a": {"b":"foo"}}'::json->'a'                  | {"b":"foo"}
+-- @
+--
+-- @since 3.1.0
+(->.) :: SqlExpr (Value (Maybe Aeson.Value))
+      -> Either Int Text
```
parsonsmatt: I think the API would be more flexible if we didn't use Haskell values directly, but instead used `SqlExpr` of these values. That's how we're doing things at Lumi right now. It's easy to go from `Text` to `SqlExpr (Value Text)` with `val`, but it's IIRC not possible to go the other way.

If we do that, though, then we need to use something other than `Either`, because (IIRC) there's no easy way to go from an eg `SqlExpr (Value Int)` to a `SqlExpr (Value (Either Int Text))`. The library sometimes defines classes like `SqlString` to allow folks to use all kinds of stuff, and I've used that before to good luck. Something like `class JsonAccessor a` with a few instances for `Int`, `Text`, `String`, `Integer`, etc would be a good way to be more flexible, at the cost of type inference. Eg, `foo ->. val 3` would give an ambiguous type warning, and `foo ->. val "hello"` would give an ambiguous warning with `OverloadedStrings` enabled.

On the other hand, we don't actually, uh, use the `SqlExpr` power *anywhere* in the Lumi codebase that I can tell. All uses of our `|->` operator have a `val "some string literal"` form.

Maybe we should have a `data JsonAccessor = JInt Int | JStr Text` that has `Num` and `IsString` instances :joy:

src/Database/Esqueleto/PostgreSQL/JSON.hs
```diff
@@ -0,0 +1,618 @@
...
+--  - | text[]  | Delete multiple key/value pairs or string elements from left operand.  | '{"a": "b", "c": "d"}'::jsonb - '{a,c}'::text[]
+--    |         | Key/value pairs are matched based on their key value.                  |
+(--.) :: SqlExpr (Value (Maybe Aeson.Value))
+      -> [Text]
+      -> SqlExpr (Value (Maybe Aeson.Value))
+(--.) value = unsafeSqlBinOp " - " value . mkTextArray
+
+-- | This operator can remove elements nested in an object.
+-- If a 'Text' is not parsable as a number when selecting in an array
+-- (even when halfway through the selection).
```
parsonsmatt: This sentence is unclear to me

src/Database/Esqueleto/PostgreSQL/JSON.hs
```diff
@@ -0,0 +1,618 @@
...
+
+
+--- ORPHAN INSTANCES ---
+--- ORPHAN INSTANCES ---
+--- ORPHAN INSTANCES ---
+
+
+-- Mainly copied over from Database.Persist.Postgresql.JSON
+-- Since we don't need anything else and adding another dependency
+-- just for these two instances is a bit overkill.
```
parsonsmatt: `esqueleto` already depends on `persistent-postgresql`. I'd *really* rather not have orphan instances potentially clashing and would definitely prefer to bring on another dependency to avoid that.

src/Database/Esqueleto/PostgreSQL/JSON.hs
```diff
@@ -0,0 +1,618 @@
...
+-- | jsonb - @since 3.1.0
+instance PersistFieldSql Aeson.Value where
+  sqlType _ = SqlOther "JSONB"
+
+badParse :: Text -> String -> Text
+badParse t = fromPersistValueParseError t . T.pack
+
+fromLeft :: (a -> b) -> Either a x -> Either b x
+fromLeft f (Left a) = Left $ f a
+fromLeft _ (Right r) = Right r
```
parsonsmatt: This is in `base` via `Data.Bifunctor` as `first`

test/PostgreSQL/Test.hs
```diff
@@ -499,6 +508,493 @@ testPostgresModule = do
...
+  -> SqlPersistT m [Entity Json]
+selectJSONwhere f = selectJSON $ where_ . f
+
+selectJSON
+  :: MonadIO m
+  => (SqlExpr (Value (Maybe A.Value)) -> SqlQuery ())
+  -> SqlPersistT m [Entity Json]
+selectJSON f = select $ from $ \v -> do
+    f $ just (v ^. JsonValue)
+    return v
```
parsonsmatt: fantastic tests :smile:

src/Database/Esqueleto/PostgreSQL/JSON.hs
Vlix: Ah, you're right P:

src/Database/Esqueleto/PostgreSQL/JSON.hs
Vlix: The use case is slim, I think. We also never use a column as the right hand operand. It's almost always taking the JSON value of a column and either filtering on it, or changing it's shape.
Though just to be able to do it, would be nice, in case someone has a specific situation where they, for example, want to filter a row based on the interaction between two columns?

Or maybe go for the data type `data JsonAccessor` and have helper functions that accept any valid value?

`numToJsonAccessor :: Num a => SqlExpr (Value a) -> SqlExpr (Value JsonAccessor)`
`stringToJsonAccessor :: IsString a => SqlExpr (Value a) -> SqlExpr (Value JsonAccessor)`

---

Also:
Any right hand operand (RHO) can be `NULL` in any of the operators. The result will just be `NULL`. If we use `SqlExpr`s, it might be good to just default to `SqlExp (Value (Maybe a))`, where a is the intended type of the RHO.

src/Database/Esqueleto/PostgreSQL/JSON.hs
Vlix: The `#-` operator removes a value nested (deep) within a JSON object/array. The RHO is a "path" to that value:

`v ^. JsonValue #-. ["user","name","last"]` would remove the "last" field in the object in the "name" field, which resides in the object that's in the "user" field of the top level object.

These examples show best what happens:

```
{"a": 3.14, "b": null}        #- ["a","b"] == {"a": 3.14, "b": null}
{"a": {"b":false}, "b": null} #- ["a","b"] == {"a": {}, "b": null}
```

The second sentence is unfinished, and should be:
"If a 'Text' is not parsable as a number when selecting in an array (even when halfway through the selection) an exception will be thrown"

src/Database/Esqueleto/PostgreSQL/JSON.hs
Vlix: The library `esqueleto:lib` doesn't.

But if you'd rather have it added in, sure, I can do that.

parsonsmatt: Good catch :sweat_smile: 

In that case I think it would *definitely* cause problems to use an orphan instance here, as most users of the library will already have `persistent-postgresql` orphan instances in scope. But then why are the tests working? I'd expect them to pick up both instances and get confused. :thinking: So maybe it's not an issue.

A consequence of adding `persistent-postgresql` as a dependency to `esqueleto` is that everyone who builds the library will need all the Postgres stuff that `pkg-config` expects for it. That's not good - mysql/sqlite users shouldn't have to install this.

It may be time to switch the DB specific modules to DB specific packages. `esqueleto-postgresql` can 100% depend on `persistent-postgresql`, and then none of this is a problem. I think that's probably the "Right" answer, but it's also the most work, so .. . .

Vlix: The reason the tests work is because the instances for `Value` are only exported from `Database.Persist.Postgresql.JSON`, and that isn't imported anywhere.

Also, the ONLY reason the test has a `persistent-postgresql` dependency, is that it uses `withPostgreslConn`. Which, I expect, is fairly easily replaced by using stuff from the `postgresql-simple` library.

I tried, and yes, importing both `Database.(Persist|Esqueleto).PostgreSQL.JSON` modules results in compiler errors...
It is a predicament, because we can't put the Aeson.Value instances in the `persistent` library, since they are specific to PostgreSQL...

We could just add a notice: __"Don't use both these methods in the same module"__?

src/Database/Esqueleto/PostgreSQL/JSON.hs
Vlix: Does this explain it? And what do you think should I add to the description to make it more obvious? *without this elaborate explanation with examples*

src/Database/Esqueleto/PostgreSQL/JSON.hs
parsonsmatt: No, that's not acceptable, because the orphans propagate. Suppose I had two modules:

```
module A where 

import Database.Esqueleto.Postgresql.JSON ()
```

```
module B where

import Database.Persist.Postgresql.JSON ()
```

`module B` and `module A` both export the orphan instance implicitly now. So any module that imports `B` and `A` and tries to use a function depending on the `instance PersistField Value` will give an error.

parsonsmatt: That's not actually right, sorry. If I define `a :: SqlPersistM Value` and `b :: SqlPersistM Value` that use the `PersistField` instances internally, then it will select the instance that was imported for those modules. Using those functions in a `module C` that imported both would be fine.

However, if we try to use the instance in any module which imports both `B` and `A` transitively, then we will get the error due to instance overlap.

Vlix: Ah, now I get why people are so against `orphan instances`.

So yeah, splitting the modules into their own packages would make the most sense, probably. Or at least to split off the `PostgreSQL.*` parts.

parsonsmatt: I need to get some broader input on what's the best release policy for this. It's a bit tricky and I'm not sure the best way to go about it.

Removing the `.Postgresql` and `.Mysql` modules is a breaking change and would signal a major bump. We can't just reexport from those modules because then we incur the dependency that we're trying to avoid. Releasing an `esqueleto-core` package without the "extras" and keeping `esqueleto` as the "kitchen sink" might work out without breaking stuff.

parsonsmatt: Maybe the use of `Value` can be hidden behind a `newtype` internally so the instance doesn't leak. But that'd allow the instance to drift between `persistent-postgresql` and `esqueleto`.

Vlix: Would that make it incompatible for tables defined with Aeson.Value (for `persistent-postgresql`) to be used by the `esqueleto` functions? I mean, you could make a coercion function `SqlExpr (Value Aeson.Value) -> SqlExpr (Value NewType)`, but the other way around (defining the table with the `NewType`) would make it impossible to use the `persistent-postgresql` functions... is that an ok trade-off? Make it clear to the user that this issue exists and that, for interoperability between `persistent-postgresql` and `esqueleto`, you should define your tables with `Aeson.Value` and not the `NewType`?

cdparks:
**tl;dr** This looks great, orphan instances bad, newtypes good, no strong opinions on arguments to `(->.)` and friends

### The longer version

Generally, I think this looks really nice, and the docs are _killer_. As far as avoiding orphan instances, we use a newtype like this:

```haskell
newtype JSONB a = JSONB { unJSONB :: a }
  deriving
    ( Generic
    , Eq
    , Foldable
    , Functor
    , Ord
    , Read
    , Show
    , Traversable
    )

instance ToJSON a => ToJSON (JSONB a) where
  toJSON (JSONB a) = toJSON a
  toEncoding (JSONB a) = toEncoding a

instance FromJSON a => FromJSON (JSONB a) where
  parseJSON = fmap JSONB . parseJSON

instance (ToJSON a, FromJSON a) => PersistFieldSql (JSONB a) where
  sqlType _ = SqlOther "JSONB"

instance (ToJSON a, FromJSON a) => PersistField (JSONB a) where
  toPersistValue = toPersistValueJSON
  fromPersistValue = fromPersistValueJSON
```

It shows up in entity declarations:

```haskell
mkPersist sqlSettings [persistLowerCase|
MathAdaptiveAssignmentAnswer sql=answers
  createdAt UTCTime
  mathQuestionId MathQuestionId
  assignmentSessionId MathAdaptiveAssignmentSessionId
  studentId StudentId
  skillId SkillId
  correctness (Percentage Double)
  durationSeconds DurationSeconds Maybe
  attempts Int Maybe
  answer (JSONB MathAnswer) Maybe -- <-------------------------- 👀 
  deriving Show Eq Generic
|]
```

And in queries:

```haskell
-- | a @> b
(@>.)
  :: SqlExpr (Value (JSONB a))
  -> SqlExpr (Value (JSONB a))
  -> SqlExpr (Value Bool)
(@>.) = unsafeSqlBinOp " @> "
infix 4 @>.

fetchSystemForSchool
  :: ... {- bunch o'constraints :) -}
  => School
  -> SqlPersistT m (Maybe system)
fetchSystemForSchool School {..} =
  selectFirstMap entityVal . from $ \systems -> do
    where_ $ systems ^. usStatesField @>. val (JSONB [schoolState])
    where_ $ systems ^. enabledField
    pure systems
```

We don't appear to have the lookup/keypath operators defined, but I think we'd be fine with either a typeclass-based approach or a concrete sum type for the key/index. In the former case, being able to use `-XTypeApplications` would go a long way towards making it friendly. Note that you can still use unstructured JSON with `JSONB Value`

src/Database/Esqueleto/PostgreSQL/JSON.hs
Vlix: I think I might have the solution. Adding the orphan instances as their separate module makes them optional to import. The `Database.Esqueleto.PostgreSQL.JSON` module doesn't need the instances, only when using them in queries. So if you don't have `persistent-postgresql` you can import the `Database.Esqueleto.PostgreSQL.JSON.Instances` module, and otherwise, just use the one from `persistent-postgresql`.

It seems to work, I tested it.

parsonsmatt:
The more I think about that, the more I like it. We also have a `newtype Jsonb` at Lumi and we use it in the manner you're describing.

Vlix:
> ```haskell
> -- | a @> b
> (@>.)
>   :: SqlExpr (Value (JSONB a))
>   -> SqlExpr (Value (JSONB a))
>   -> SqlExpr (Value Bool)
> (@>.) = unsafeSqlBinOp " @> "
> infix 4 @>.
> ```

It's a bit too restrictive in that both the LHO and RHO are `SqlExpr (Value (JSONB a))`, but otherwise seems fine on first glance. Though I think this makes it more difficult to chain operators...

cdparks:
> Though I think this makes it more difficult to chain operators

Yeah, this signature can definitely be loosened up, that's a product of us only using this in a few places

Vlix:
In my opinion, using `Maybe Aeson.Value` is the most honest representation, since most operators can arbitrarily change the structure of the JSON, and having that reflected in the types makes it obvious you need to be careful.
Then again, if you keep to your own invariants and make sure no garbage JSON enters your tables, it should be fine to be more specific (using `JSONB OwnType`). I also immediately think about how I would make that work with the `safe-json` package I made. If it's possible to make that easy to use, that'd be great.

cdparks:
> In my opinion, using `Maybe Aeson.Value` is the most honest representation

It'd be possible to have the operators do this and _internally_ use a newtype to handle instance selection without an orphan:

```haskell
newtype StoreJSON = StoreJSON { unStoreJSON :: Value }
  deriving (...)

instance PersistField StoreJSON where
  ...

(->>.)
  :: SqlExpr (Value (Maybe Aeson.Value))
  -> Either Int Text
  -> SqlExpr (Value (Maybe Text))
(->>.) value (Right txt) = unsafeSqlBinOp " ->> " (StoreJSON value) $ val txt
(->>.) value (Left i)    = unsafeSqlBinOp " ->> " (StoreJSON value) $ val i
```

That said, I think `Maybe (JSONB a)` (or some equivalent `newtype`) gets you the best of both worlds because you can always drop down to `Maybe (JSONB Aeson.Value)` if you need to. I worry that privileging `Maybe Aeson.Value` would lead people to put `Aeson.Value` in their entity declarations. `JSONB Something` says "this thing is structured and merely uses JSON for serialization". `JSONB Value` says "this is a semi-or-un-structured blob, be careful".

parsonsmatt:
I think there's a lot of value in the "raw" `Value` API. Safer APIs can be built on top of that relatively easily. `SqlExpr (Value (Jsonb Value)) -> SqlExpr (Value (Jsonb yourType))` is just an `veryUnsafeCoerceSqlExpr` away :joy: 

Perhaps with `Generic`s and `OverloadedLabels` we could have an API like `(rec :: SqlExpr (Value (Jsonb MyType))) ->. #myTypeSomeField :: SqlExpr (Value (Jsonb SomeFieldType))`, given something like

```haskell
data MyType = MyType { myTypeSomeField :: SomeField }

data SomeField = SomeField { someFieldThing :: Int }
```

Not really relevant to this PR, but I would like to consider APIs that we can extend safely later on without invasive refactoring upstream.

`SqlExpr (Maybe (Jsonb Value)) -> Text -> SqlExpr (Maybe (Jsonb Value))` is basically the same API in this PR, but it doesn't require an orphan instance, doesn't require dancing around package releases, doesn't break anything, and easily integrates with the `Jsonb CustomType` with some casts.

Using `(ToJSON a, FromJSON a, ToJSON b, FromJSON b) => SqlExpr (Maybe (Jsonb a)) -> Text -> SqlExpr (Maybe (Jsonb b))` can easily be monomorphized to the above signature. I don't think that this signature would have awful type inference properties - the `SqlExpr` constructors are either going to be projecting a field from an entity (with a "known" result type) or using `val` with a "known" argument type. The result type may not be known, though, and that may need annotation.

However, the signature we use at Lumi is not restricted:

```haskell
(|->?) :: SqlExpr (Value (Maybe (Jsonb a))) -> SqlExpr (Value Text) -> SqlExpr (Value (Maybe (Jsonb b)))
(|->?) = ((unsafeSqlFunction "jsonb_object_field") .) . (,)
```

This doesn't present any problems chaining. Maybe the constraints are unnecessary, except for when trying to deserialize the value.

parsonsmatt:
> In my opinion, using Maybe Aeson.Value is the most honest representation, since most operators can arbitrarily change the structure of the JSON, and having that reflected in the types makes it obvious you need to be careful.

This is definitely true! But Esqueleto isn't trying to be a 100% honest representation of SQL - it tries to give best-of-both-worlds to SQL and Haskell by allowing you to use Haskell types as much as possible. The most "honest" representation for many `Key entity` type is `Int64` but we have a richer language of types than SQL can support here.

Vlix:
That is very true. I'll try to add a `newtype` and see if any tests go haywire.

I'm wondering if the type inference will be fine with, e.g.:

```haskell
v ^. JsonValue ->. Right "test" -. Left 1 @>. just (val $ Number 3)
```

src/Database/Esqueleto/PostgreSQL/JSON.hs
parsonsmatt: Finishing the second sentence clears it right up for me, thanks! :smile:

Vlix:
Ok, seems like it wasn't too much trouble. Added some conveniences as well:

- type synonym: `JSONExpr`
- function: `jsonbVal`

Still not sure if we want to keep the raw `Text / [Text] / Either Int Text`...
Putting the `Text` in `SqlExpr (Value (Maybe Text))` isn't such a big deal, but the other ones are a bit iffy.
The `[Text]` would only work out of a column if that column was a `text[]` (PostgreSQL text array type). And the 'Either Int Text' would have to be solved with, I think, a type class in the same style as `SqlString`?

Vlix:
@eborden
Do you have an opinion about the RHOs that aren't in `SqlExpr (Value a)`?

src/Database/Esqueleto/PostgreSQL/JSON.hs
Vlix: Implemented JSONB newtype for `PersistField(Sql)` instances and also type polymorphism, as long as the provided type has `From-/ToJSON` instances.

eborden:
@Vlix the general pattern in `esqueleto` up to this point has been to use open world assumptions and define a typeclass like [`SqlString`](https://www.stackage.org/haddock/nightly-2019-08-01/esqueleto-3.0.0/Database-Esqueleto.html#t:SqlString). Following that pattern would align your RHO values with existing convention. I don't know if I'm particularly compelled by that option, but in some cases it does open up your strict structure to more options like `Vector Text` in place of `[Text]`.

parsonsmatt:
My only issue with the polymorphic `class JsonAccessor` style is that I anticipate using literals with this somewhat often - indeed every use case of the `->.` operator at Lumi uses a literal `Text`, and trying to use these would require a type annotation at every use site.

That's why I'm more OK with defining a somewhat dodgy

```haskell
data JsonAccessor
  = JsonInt Integer
  | JsonText Text

instance Num JsonAccessor where
  fromInteger = JsonInt
  (+) = error "forgive me"

instance IsString JsonAccessor where
  fromString = JsonText . Text.pack
```

If you want the nice syntax, you can opt-in via numeric literals. If you want to use the constructors, they're right there - `JsonText "hello"` is IMO preferable to `Right "hello"`. It would be a little annoying if you already have an `Integer` that you need to wrap it in a `fromInteger` or `JsonInt` call. Idk. We don't actually have any uses of the array access functions - we mostly tend to do a `CROSS JOIN` on array elements and treat them as tables.

eborden:
Agreed that these will mostly be literal values.

cdparks:
The concrete sum type works for me. I'd probably call them `JsonIndex` and `JsonKey`, but that's neither here nor there. Every version of the polymorphic version I can dream up feels like a lot of machinery for not much gain.

Vlix:
I've taken your suggestions and implemented the `JSONAccessor` data type. Even with the dodgy `Num` instance. 🙃

Do you guys think anything's still missing? Or anything can still be improved upon?

src/Database/Esqueleto/PostgreSQL/JSON/Instances.hs
```diff
@@ -0,0 +1,118 @@
...
+    , Foldable
+    , Functor
+    , Ord
+    , Read
+    , Show
+    , Traversable
+    )
+
+-- | 'SqlExpr' of a NULL-able 'JSONB' value.
+type JSONExpr a = SqlExpr (Value (Maybe (JSONB a)))
```
paf31: One small point - I understand why the `Maybe` is necessary here, but I find it a bit strange that `JSONExpr` is the name of the synonym for nullable JSONB values. 

When I read `(->.) :: JSONExpr a -> JSONAccessor -> JSONExpr b`, I don't see any immediate indication that anything there is nullable, and I have to look at the type synonym to see it.

I'd suggest changing the name to indicate the `Maybe` in there - perhaps:

```haskell
-- | 'SqlExpr' of a NULL-able 'JSONB' value.
-- Note: The null here is a SQL null, not a JSON null.
type JSONMaybeExpr a = SqlExpr (Value (Maybe (JSONB a))) -- name isn't great
```

On a related note, I don't see any way to operate on _non-nullable_ JSONB values. Is that by design? Presumably it would be quite hard to provide an API which dealt with both without a lot of duplication, and I suppose I can always use `just` on any exprs from non-nullable columns.

Please don't consider this to be blocking at all.

paf31: On the subject of changing the name, maybe it would also be good to add the `B` to indicate this in `JSONB`, in case there is any chance of `JSON` types being added later.

Vlix: I'd say, you'll notice when you use it, and you can always look at the type synonym.
If someone has a good suggestion on how to make this clearer, I can easily integrate it.

Vlix: > On the subject of changing the name, maybe it would also be good to add the `B` to indicate this in `JSONB`, in case there is any chance of `JSON` types being added later.

This is probably a good idea, so I'll adjust that.

parsonsmatt:
Thanks for the awesome PR @Vlix !

Issue #6, opened 2016-12-15
Convenience function for dumping SQL


bitemyapp:
This is a touch annoying at the moment.

Optional prettified rendering would be nice too.

PR #133, opened 2019-08-28
Render queries as Text

parsonsmatt:


bitemyapp:
I've needed this semi-frequently. Let me know if the build needs cleaned up. @parsonsmatt

Vlix:
Seems like the same issue I had when trying to make it work locally vs. Travis?
Since I had to up the PSQL version to 10 (for specific JSONB operators), the PSQL port is now 5433, instead of 5432. (or the other way around?) but somewhere that might have to be adjusted before it'll build on Travis.

When working on the JSONB PR locally it would just work with 5432 (or 5433, can't remember), but then Travis complained if it wasn't set to the other before pushing.


From the failed Travis build:

```
$ echo "CREATE USER esqutest WITH PASSWORD 'esqutest';" | psql postgres
psql: could not connect to server: No such file or directory
	Is the server running locally and accepting
	connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5433"?
The command "echo "CREATE USER esqutest WITH PASSWORD 'esqutest';" | psql postgres" failed and exited with 2 during .
```

bitemyapp:
I believe this fixes #6

PR #137, opened 2019-09-06
Fix code example of LeftOuterJoin and ?.

rnons:


parsonsmatt:
I'm going to merge this despite the failing CI because the failing CI is extraneous and this is a doc-level change.

Thanks for the contribution!

PR #138, opened 2019-09-17
Fix LeftOuterJoin example in README.

K0Te:
LeftOuterJoin example in README contains typo and doesn't type check.
This PR fixes it.


parsonsmatt:
Thanks!

Issue #56, opened 2017-10-08
upsert/upsertBy Support


FintanH:
Had to fall back to Persistent when using `upsertBy`. I believe `upsertBy` was imported from esqueleto but I had to use the following to type check:

```
import qualified Database.Persist as P

upsertFoo foo = upsertBy (UniqueName $ name foo) foo [FooBar P.=. bar d]
```

where I'd imagine the syntax being:

```
import Database.Esqueleto

upsertFoo foo = upsertBy (UniqueName $ name foo) foo [foo ^. FooBar =. bar d]
```

PR #147, opened 2019-09-26
removed persist's PersistUniqueWrite and add upsertBy and upsert

JoseD92:


JoseD92:
@bitemyapp I add upsertBy and upsert functions that use the types of esqueleto, with this, the expression on #56 can  be written as `upsertFoo foo = upsertBy (UniqueName $ name foo) foo [FooBar =. bar]` a bit different than requested but makes more sense this way.

I had to remove `PersistUniqueWrite` class imported from persist, because the class was defining the types it had to go, but I have not re implemented deleteBy, insertUnique and putMany but should not be difficult to do. Either we rename esqueleto's upsertBy and upsert so they don't collide with persist's or we create our own version of `PersistUniqueWrite`, I personally prefer the last option that way we can specialize the functions for each backend. The implementation I made for upsertBy and upsert work for postgres but not sure for other backends, that is why I say we should make our own  `PersistUniqueWrite` but want to know you opinion.

After deciding what to do with upsertBy and upsert I will add some test, so don't merge anything yet.

bitemyapp:
This PR mingles reformatting with new functionality, closing. Please make a new PR with only the changes necessary for `upsert`/`upsertBy`.

PR #148, opened 2019-09-27
WIP: add EsqueletoUpsert class and SqlBackend instance

### LINKED ISSUE #56: upsert/upsertBy Support

JoseD92:
fixes #56

### DIFF
--- a/esqueleto.cabal
+++ b/esqueleto.cabal
@@ -49,6 +49,7 @@ library
     , bytestring
     , conduit >=1.3
     , monad-logger
+    , mtl
     , persistent >=2.10.0 && <2.11
     , resourcet >=1.2
     , tagged >=0.2
--- a/src/Database/Esqueleto.hs
+++ b/src/Database/Esqueleto.hs
@@ -100,6 +100,7 @@ module Database.Esqueleto
     -- * Helpers
   , valkey
   , valJ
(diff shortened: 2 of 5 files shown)


JoseD92:
I hide Database.Persist.upsert and Database.Persist.upsertBy and add a class EsqueletoUpsert, for backends that can do upsert, the basic behavior is doing a get by the unique field, update if exist and then select the field and return it.

I also add a instance EsqueletoUpsert SqlBackend so that it uses connUpsertSql from the SqlBackend if any defined, right now only postgres uses that, that way the 3 sql operations of the default behavior become one.

bitemyapp:
`upsert`/`upsertBy` functionality is really just for PostgreSQL so we won't be using a type class to cover implementations for different databases. I'm not keen on proliferating the use of type classes to handle this either. I've been comfortable so far with having database specific modules.

PR #152, opened 2019-10-17
[WIP] sub_select is nullable

### LINKED ISSUE #151: RFC: `sub_select` should have a nullable result type by default
tl;dr: `sub_select` should have the type `sub_select :: SqlQuery (SqlExpr (Value a)) -> SqlExpr (Value (Maybe a))`.

We use `sub_select` extensively in the work codebase, and it is a common source of runtime errors. The general pattern is this:

```haskell
bad :: SqlPersistM [(Entity Foo, XPropType)]
bad = 
  select $
  from $ \foo -> do 
  where_ (foo ^. FooProperty ==. val 3)
  pure (foo, sub_select (from_ \x -> where_ (p x) >> limit 1 >> pure (x ^. XProp))
```

`sub_select` has the type `sub_select :: SqlQuery (SqlExpr (Value a)) -> SqlExpr (Value a)`. Because the result type is `Value a`, the relevant marshalling code assumes that it cannot be `NULL`. However, it totally can be `NULL`, for a huge amount of the potential uses - possibly even the majority or default case.

A problem is that it is going to be quite difficult to statically determine whether or not it will return null. `sub_select $ from $ \x -> where_ (p x) >> limit 1 >> pure (x ^. XProp)` will obviously return `NULL` if the query returns 0 results, and `persistent` will fail to parse a `Nothing`. The work-around is to use `just`.

However, `sub_select $ from $ \x -> where_ (p x) >> pure (sum (x ^. XProp))` will return `0` for zero rows, so it doesn't *need* to be `Maybe`. Likewise, `sub_select $ from $ \x -> where_ (p x) >> pure countRows` will also return a non-null result.

I propose that `sub_select` be made nullable by default, along with another function that can dismiss the `Maybe` when the user knows exactly what they're doing and it'll be fine.

parsonsmatt:
Fixes #151 

I'm going to apply this on the work codebase and see what comes up.

Alternatively, we could deprecate `sub_select` and the deprecation notice would point to `subSelect` with this type and `subSelectUnsafe` with the same type.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,12 @@
+Unreleased (3.2.0)
+=======
+
+- @parsonsmatt
+  - [#152](https://github.com/bitemyapp/esqueleto/pull/152): Changed the type of
+    `sub_select` to return a `Maybe`. This fixes a potential runtime error where
+    the subquery returns zero rows, resulting in a `NULL` for the expression
+    that `persistent` cannot parse.
+
 Unreleased (3.1.1)
 ========
 
--- a/src/Database/Esqueleto/Internal/Internal.hs
+++ b/src/Database/Esqueleto/Internal/Internal.hs
@@ -350,7 +350,10 @@ locking kind = Q $ W.tell mempty { sdLockingClause = Monoid.Last (Just kind) }
 sub_select :: PersistField a => SqlQuery (SqlExpr (Value a)) -> SqlExpr (Value (Maybe a))
 sub_select         = just . sub SELECT
 
--- |
+-- | Promise that the value you have is /definitely/ not @NULL@, by some
+-- guarantee that doesn't exist in esqueleto's types.
+--
+-- @since 3.2.0
 unsafePromiseNotNull
   :: SqlExpr (Value (Maybe a))
   -> SqlExpr (Value a)
--- a/src/Database/Esqueleto/Internal/Internal.hs
+++ b/src/Database/Esqueleto/Internal/Internal.hs
@@ -342,8 +342,19 @@ locking kind = Q $ W.tell mempty { sdLockingClause = Monoid.Last (Just kind) }
 -- | Execute a subquery @SELECT@ in an SqlExpression.  Returns a
 -- simple value so should be used only when the @SELECT@ query
 -- is guaranteed to return just one row.
-sub_select :: PersistField a => SqlQuery (SqlExpr (Value a)) -> SqlExpr (Value a)
-sub_select         = sub SELECT
+--
+-- In 3.2.0, this was updated to have a @Maybe@ in the return type. This change
+-- is due to the common error of using it with queries that might return @NULL@.
+-- Call the function 'unsafePromiseNotNull' if you can guarantee that the value
+-- will return exactly one row.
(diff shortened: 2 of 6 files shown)


src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -579,8 +590,11 @@ castString = veryUnsafeCoerceSqlExprValue
 
 -- | Execute a subquery @SELECT@ in an SqlExpression.  Returns a
 -- list of values.
-subList_select :: PersistField a => SqlQuery (SqlExpr (Value a)) -> SqlExpr (ValueList a)
-subList_select         = EList . sub_select
+subList_select
+  :: PersistField a
+  => SqlQuery (SqlExpr (Value a))
+  -> SqlExpr (ValueList a)
+subList_select         = EList . unsafePromiseNotNull . sub_select
```
parsonsmatt: I *think* this should do a `NULL -> []` transformation but I am not sure. Will need to write a test against it.

changelog.md
```diff
@@ -1,3 +1,12 @@
+Unreleased (3.2.0)
+=======
+
+- @parsonsmatt
+  - [#152](https://github.com/bitemyapp/esqueleto/pull/152): Changed the type of
+    `sub_select` to return a `Maybe`. This fixes a potential runtime error where
+    the subquery returns zero rows, resulting in a `NULL` for the expression
+    that `persistent` cannot parse.
+
 Unreleased (3.1.1)
```
parsonsmatt: I'll make a 3.1.1 release soon.

parsonsmatt:
Call for commentary: @cdparx @eborden @andrewthad @dsh 

Can test this out with this entry in your extra-deps:

```yaml
- git: https://github.com/bitemyapp/esqueleto
  commit: 7e92f751bd607766e75baf7d710edf8d605c4de6
```

parsonsmatt:
I've been working on this at the Lumi codebase for a bit today, and the pattern I'm noticing is this:

- It's used with `max_`, `sum_`, etc which are safe, and so I use a `joinV . sub_select`
- It's used with `countRows`, which is safe, so I use `unsafePromiseNotNull . sub_select`
- It's used with a query, which we've already done `pure (just result)`, and I can refactor to mere `pure result`.
- It's used with something like `sub_select $ from $ \p -> pure (coalesceDefault [f p] (val x)`, which can be refactored to `coalesceDefault [sub-select $ from $ \p -> pure (f p)] (val x)` to regain safety.

So I think the better thing to do here is provide the deprecation warning, which should point to replacement functions:

- `subSelect :: SqlQuery (SqlExpr (Value a)) -> SqlExpr (Value (Maybe a))` as the "safe" default. Maybe it even throws a `limit 1` on there, just to be *extra* safe.
- `subSelectMaybe :: SqlQuery (SqlExpr (Value (Maybe a))) -> SqlExpr (Value (Maybe a))`, which is just `joinV . subSelect`
- `subSelectUnsafe :: SqlQuey (SqlExpr (Value a)) -> SqlExpr (Value a)` which does no checks, but is safe to use with with eg `countRows`.

parsonsmatt:
Closing in favor of #153

Issue #86, opened 2018-04-10
Support `replaceSelect`


tysonzero:
It would be nice to be able to use `insertSelect` but where rows with overlapping unique constraints were replaced.

parsonsmatt:
This is -- unfortunately -- database specific. MySQL and Postgres have different syntax for this functionality. So we'd need to gate that behind the `.MySQL` and `.Postgresql` modules.

JoseD92:
Based on this comments I came with a solution for postgres in a function named `insertSelectWithConflict`, firm `(MonadIO m, PersistEntity val) => Unique val -> SqlQuery (SqlExpr (Insertion val))  -> (SqlExpr (Entity val) -> SqlExpr (Entity val) -> [SqlExpr (Update val)]) -> SqlWriteT m ()`, it works similar to `insertSelect` but allows solving conflicts on a single constraint when doing `insertSelect`, @tysonzero problem could be solved using this function, but he will need to specify updating all the values using the excluded values, I though I could code a `replaceSelect` based on `insertSelectWithConflict` but turns out there is no way to get the lenses for each field of a datatype on a single list from persist, so I could not generate the updates list automatically, and the other solution was building all from strings that may introduce some error along the way, given the `insertSelectWithConflict` already accomplishes the task with a small manual work and that it allows great flexibility I think is a good solution for this issue.

PR #155, opened 2019-10-22
Insert Select With Conflict for postgres

JoseD92:
Based on #86 I came with insertSelectWithConflict, it is like insertSelect but allows to perform an update if some row conflicts with the new addition on a given constraint. There is no way in postgres (that I could find) of responding to any possible conflict, so a specific one needs to be indicated with a unique.

### DIFF
--- a/src/Database/Esqueleto/PostgreSQL.hs
+++ b/src/Database/Esqueleto/PostgreSQL.hs
@@ -302,6 +302,8 @@ insertSelectWithConflict :: forall a m val. (
 insertSelectWithConflict unique query = void . insertSelectWithConflictCount unique query
 
 -- | Same as 'insertSelectWithConflict' but returns the number of rows affected.
+--
+-- @since 3.1.3
 insertSelectWithConflictCount :: forall a val m. (
     FinalResult a,
     KnowResult a ~ (Unique val), 
--- a/src/Database/Esqueleto/PostgreSQL.hs
+++ b/src/Database/Esqueleto/PostgreSQL.hs
@@ -242,6 +242,11 @@ instance FinalResult (Unique val) where
 instance (FinalResult b) => FinalResult (a -> b) where
   finalR f = finalR (f undefined)
 
+-- | Convert a constructor for a 'Unique' key on a record to the 'UniqueDef' that defines it. You 
+-- can supply just the constructor itself, or a value of the type - the library is capable of figuring 
+-- it out from there.
+--
+-- @since 3.1.3
 toUniqueDef :: forall a val. (KnowResult a ~ (Unique val), PersistEntity val,FinalResult a) => 
   a -> UniqueDef
 toUniqueDef uniqueConstructor = uniqueDef
--- a/src/Database/Esqueleto/PostgreSQL.hs
+++ b/src/Database/Esqueleto/PostgreSQL.hs
@@ -2,9 +2,7 @@
 {-# LANGUAGE OverloadedStrings
            , GADTs, CPP, Rank2Types
            , ScopedTypeVariables
-           , FlexibleInstances
  #-}
- {-# LANGUAGE TypeFamilies #-}
 -- | This module contain PostgreSQL-specific functions.
 --
 -- /Since: 2.2.8/
@@ -27,7 +25,6 @@ module Database.Esqueleto.PostgreSQL
   , insertSelectWithConflictCount
   -- * Internal
   , unsafeSqlAggregateFunction
-  , toUniqueDef
   ) where
 
 #if __GLASGOW_HASKELL__ < 804
@@ -40,7 +37,8 @@ import           Database.Esqueleto.Internal.PersistentImport hiding (upsert, up
 import           Database.Esqueleto.Internal.Sql
 import           Database.Esqueleto.Internal.Internal         (EsqueletoError(..), CompositeKeyError(..), 
                                                               UnexpectedCaseError(..), SetClause, Ident(..),
-                                                              uncommas)
+                                                              uncommas, FinalResult(..), toUniqueDef,
+                                                              KnowResult, renderUpdates)
 import           Database.Persist.Class                       (OnlyOneUniqueKey)
 import           Data.List.NonEmpty                           ( NonEmpty( (:|) ) )
 import           Data.Int                                     (Int64)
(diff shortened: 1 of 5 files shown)


src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -203,14 +209,81 @@ upsertBy uniqueKey record updates = do
...
+--   (\current excluded ->
+--     [FooNum =. (current ^. FooNum) +. (excluded ^. FooNum)]
+--   )
+-- @
+--
+-- Inserts to table Foo all Bar.num values and in case of conflict SomeFooUnique, 
+-- the conflicting value is updated to the current plus the excluded.
+insertSelectWithConflict :: (MonadIO m, PersistEntity val) => 
+  Unique val
+  -- ^ Uniqueness constraint, this is used just to get the name of the postgres constraint, the value(s) is(are) never used, so if you have a unique "MyUnique 0", "MyUnique undefined" will work as well.
```
parsonsmatt: I've used a trick like this before that matches on the final result type of a value - see [`KnowResult`](https://hackage.haskell.org/package/persistent-documentation-0.1.0.1/docs/src/Database.Persist.Documentation.Internal.html#KnowResult) here which allows you to only provide the data constructor. With another type class, you can even pass in the required `undefined`s automatically so that the user is not concerned with this.

test/PostgreSQL/Test.hs
```diff
@@ -978,6 +978,50 @@ testUpsert =
...
+            from $ \p -> return $ OneUnique <# val "test" <&> (p ^. PersonFavNum)
+          )
+          (\current excluded -> [OneUniqueValue =. val 4 +. (current ^. OneUniqueValue) +. (excluded ^. OneUniqueValue)])
+        uniques2 <- select $ from $ \u -> return u
+        liftIO $ n1 `shouldBe` 3
+        liftIO $ n2 `shouldBe` 3
+        let test = map (OneUnique "test" . personFavNum) [p1,p2,p3]
+            test2 = map (OneUnique "test" . (+4) . (*2) . personFavNum) [p1,p2,p3]
+        liftIO $ map entityVal uniques1 `shouldBe` test
+        liftIO $ map entityVal uniques2 `shouldBe` test2
```
parsonsmatt: Very nice, thanks for the tests!

src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -203,14 +209,81 @@ upsertBy uniqueKey record updates = do
...
+  where
+    proxy :: Proxy val
+    proxy = Proxy
+    updates = conflictQuery entCurrent entExcluded
+    combine (tlb1,vals1) (tlb2,vals2) = (builderToText (tlb1 `mappend` tlb2), vals1 ++ vals2)
+    entExcluded = EEntity $ I "excluded"
+    tableName = unDBName . entityDB . entityDef
+    entCurrent = EEntity $ I (tableName proxy)
+    -- there must be a better way to get the constrain name from a unique, make this not a list search
+    uniqueDef = head . filter ((==) (persistUniqueToFieldNames unique) . uniqueFields) . entityUniques . entityDef $ proxy
```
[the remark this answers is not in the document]
JoseD92: smart guy, I always forget `find` exists, but it keeps being a list search, but I think it should not be too difficult to add `toUniqueDef` on persist, though I will have to look at the code to be sure.

Also, the thing with the undefined, I would also like it to be automatic, but the Unique creation function may have any amount of fields, when defining a unique in persist we can:

```
Foo
    num1 Int
    num2 Int
    num3 Int
    UniqueFooNum1 num1
    UniqueFooNum23 num2 num3
    deriving Eq Show
```

so I can't just change the firm to `(a -> Unique val) -> SqlQuery (SqlExpr (Insertion val)) -> (SqlExpr (Entity val) -> SqlExpr (Entity val) -> [SqlExpr (Update val)]) -> SqlWriteT m Int64` because I would not be able to use `UniqueFooNum23` on this new function, if there is a way receive a variable parameter function and then fill all parameters with undefined I will add it.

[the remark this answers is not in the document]
parsonsmatt: I linked to `KnowResult` in another comment. You can implement a type class that will "fill in" the unused values with `undefined` also. It's a bit of a trick of type class programming so if you'd like help on that please let me know :)

src/Database/Esqueleto/PostgreSQL.hs
JoseD92: Thanks! nice recommendation, I have not use type families much, but seeing `KnowResult` did help me avoid the undefined, though I think I should move `KnowResult` to another place (I copy `KnowResult`  to Database.Esqueleto.PostgreSQL), any recommendations on where to place it and if it is possible to make the firm for insertSelectWithConflict prettier?

src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -224,14 +227,52 @@ renderUpdates conn = uncommas' . concatMap renderUpdate
...
+  KnowResult a = a
+
+class FinalResult a where
+  finalR :: a -> KnowResult a
+
+instance FinalResult (Unique val) where
+  finalR = id
+
+instance (FinalResult b) => FinalResult (a -> b) where
+  finalR f = finalR (f undefined)
```
parsonsmatt: Well done!!

parsonsmatt:
Is this still WIP?

src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -199,18 +208,130 @@ upsertBy uniqueKey record updates = do
...
-        where
-          mk :: SqlExpr (Value ()) -> [(TLB.Builder, [PersistValue])]
-          mk (ERaw _ f)        = [f info]
-          mk (ECompositeKey _) = throw (CompositeKeyErr MakeSetError) -- FIXME
-          renderUpdate :: SqlExpr (Update val) -> [(TLB.Builder, [PersistValue])]
-          renderUpdate (ESet f) = mk (f undefined) -- second parameter of f is always unused
-          info = (projectBackend conn, initialIdentState)
\ No newline at end of file
+
+-- | Render postgres updates to be use in a SET clause.
```
parsonsmatt:
```suggestion
-- | Render postgres updates to be use in a SET clause.
-- 
-- @since 3.1.3
```

Please add an `@since` annotation with the version number that this gets.

src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -199,18 +208,130 @@ upsertBy uniqueKey record updates = do
...
+class FinalResult a where
+  finalR :: a -> KnowResult a
+
+instance FinalResult (Unique val) where
+  finalR = id
+
+instance (FinalResult b) => FinalResult (a -> b) where
+  finalR f = finalR (f undefined)
+
+toUniqueDef :: forall a val. (KnowResult a ~ (Unique val), PersistEntity val,FinalResult a) =>
```
parsonsmatt:
```suggestion
-- | Convert a constructor for a 'Unique' key on a record to the 'UniqueDef' that defines it. You 
-- can supply just the constructor itself, or a value of the type - the library is capable of figuring 
-- it out from there.
--
-- @since 3.1.3
toUniqueDef :: forall a val. (KnowResult a ~ (Unique val), PersistEntity val,FinalResult a) => 
```

Since this is exported, please add a documentation comment along with an `@since` annotation for the version number.

src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -199,18 +208,130 @@ upsertBy uniqueKey record updates = do
...
+--   (from $ \b ->
+--     return $ Foo <# (b ^. BarNum)
+--   )
+--   (\current excluded ->
+--     [FooNum =. (current ^. FooNum) +. (excluded ^. FooNum)]
+--   )
+-- @
+--
+-- Inserts to table Foo all Bar.num values and in case of conflict SomeFooUnique, 
+-- the conflicting value is updated to the current plus the excluded.
```
parsonsmatt:
```suggestion
-- the conflicting value is updated to the current plus the excluded.
--
-- @since 3.1.3
```

src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -199,18 +208,130 @@ upsertBy uniqueKey record updates = do
...
+  a
+  -- ^ Unique constructor or a unique, this is used just to get the name of the postgres constraint, the value(s) is(are) never used, so if you have a unique "MyUnique 0", "MyUnique undefined" would work as well.
+  -> SqlQuery (SqlExpr (Insertion val)) 
+  -- ^ Insert query.
+  -> (SqlExpr (Entity val) -> SqlExpr (Entity val) -> [SqlExpr (Update val)])
+  -- ^ A list of updates to be applied in case of the constraint being violated. The expression takes the current and excluded value to produce the updates.
+  -> SqlWriteT m ()
+insertSelectWithConflict unique query = void . insertSelectWithConflictCount unique query
+
+-- | Same as 'insertSelectWithConflict' but returns the number of rows affected.
```
parsonsmatt:
```suggestion
-- | Same as 'insertSelectWithConflict' but returns the number of rows affected.
--
-- @since 3.1.3
```

parsonsmatt:
Once CI is green I'll merge. Thanks @JoseD92 !!

parsonsmatt:
3.1.3 is released. Thanks again!

Issue #49, opened 2017-08-19
Bad (not reverse of model introduction) JOIN order causes runtime error


bitemyapp:
This needs to be documented so that we understand precisely what causes it and what can be done to fix it.

FintanH:
I found the author explaining why the reverse order is needed in the documentation in `Language.hs`.

> The order is reversed in order to improve composability. For example, consider query1 and query2 below:
> 
> ``` Haskell
> let query1 =
>       from $ \(foo `InnerJoin` bar) -> do
>         on (foo ^. FooId ==. bar ^. BarFooId)
>     query2 =
>       from $ \(mbaz `LeftOuterJoin` quux) -> do
>         return (mbaz ?. BazName, quux)
>     test1 =      (,) <$> query1 <*> query2
>     test2 = flip (,) <$> query2 <*> query1
> ```
> If the order was not reversed, then test2 would be broken: query1's on would refer to query2's LeftOuterJoin.

I'm trying to tease out why this breaks with his example, looking at the types, but maybe you can spot the reason quicker than I can.

qrilka:
While porting significant amount of queries (BTW @bitemyapp you've seen many of them I think :) )  from `mysql-simple` to `persistent` + `esqeleto` this have bitten me a bit - it just gives SQL error, not even exception from the library.
@FintanH it breaks because joins and "on" conditions get mismatched, i.e. condition from `query1` gets used for `LeftOuterJoin` between `mbaz` and `quux` while it talks about `foo` and `bar`.

MaxGabriel:
I know this isn't getting fixed in the current design, but has anyone come up with any workaround or way to programmatically check this by any chance? It's been the cause of two bugs for us

bitemyapp:
@MaxGabriel The closest thing to a workaround would be [dovetail](https://github.com/frontrowed/dovetail) but I am pretty sure Front Row Ed no longer uses it.

Otherwise, I need time to work on [bluetick](https://github.com/bitemyapp/bluetick).

parsonsmatt:
I'm exploring this to see if there's something that can be done to fix this up.

First, let's look at the offending function, `on`:

```haskell
on :: SqlExpr (Value Bool) -> SqlQuery ()
on expr = Q $ W.tell mempty { sdFromClause = [OnClause expr] }
```

The definition of `SqlQuery`:

```haskell
-- | SQL backend for @esqueleto@ using 'SqlPersistT'.
newtype SqlQuery a =
[...]

PR #156, opened 2019-10-23
Fix the On Clause Ordering issue

parsonsmatt:
This PR tracks my attempt to fix the on clause ordering issue.

[Previous discussion here](https://github.com/bitemyapp/esqueleto/issues/49#issuecomment-545223824).

TLDR on implementation plan:

- Render the `OnClause (SqlExpr (Value Bool))` into `Text` (lmao)
- Parse the `Text` and find table identifiers - we pretty much have a guarnatee that they'll have the grammar:
    ```
    escape-char [character] escape-char . escape-char [character] escape-char
    ```
    which should make it relatively straightforward to write a parser that will return all the `Ident`s for the `OnClause`.
- Once we have the identifiers associated with the `OnClause`, we can be more selective about finding the appropriate match. The `FromClause` that gets constructed is a binary tree with nodes of `FromClause lhs JoinKind rhs (Maybe OnClause)` and leaves of `FromStart Ident EntityDef`. The binary tree gets associated to the left (so it's really more like a list).
- The right-most `FromStart` is meant to be the base of the join, so it's name does not need to be in the `OnClause` for the top-level `FromJoin`.

# Longer form implementation plan

```haskell
from $ \(a `InnerJoin` b `InnerJoin` c) -> do
on (c ^. CBId ==. b ^. BId)
on (b ^. BAId ==. a ^. AId)
```

This will generate a `[FromClause]` like this:

```haskell
[ FromClause ...
, OnClause ...
, OnClause ...
]
```

That `FromClause` will be shaped like this:

```
                 FromJoin 
                /        \
            FromJoin      FromStart A
           /        \
FromStart C         FromStart B
```

You might look at this and think, "Ah, a degenerate tree - let's just make it a list!" Unfortunately, Esqueleto supports arbitrary nesting of joins, in part to make composability better.

We'd start at the top: `FromJoin lhs rhs`. We want the left-most leaf of `rhs` and the right-most leaf of `lhs`. This is easy for the first `FromJoin` because the `rhs` is a `FromStart`. It has the table identifier `a`. For the next, we get into the `lhs`, which is another `FromJoin`. We descend to the right, and we get to a `FromStart` with an identifier `b`. 

Now, we'll go and render/parse our OnClauses. We take our first `OnClause`, and we render it into `"c"."b_id" = "b"."b_id"`. We can parse out the two table identifiers - `"c"` and `"b"`. This doesn't match. So we render/parse the second one, which gives us `"b"."a_id" = "a"."id"` and the idents `a` and `b`. This is a match. So we can associate *this* `OnClause` with the top-level `FromJoin`.

Now, we need to find the "next" `FromJoin`. The initial case will always be `FromJoin rhs FromStart`. It gets more complex here. The "next" `FromJoin` is going to be a `FromJoin` that has a `FromStart` as the RHS, and should be the rightmost one. In this simple case, we'd descend into the second `FromJoin` and check the RHS - it's a `FromStart`, so we're good. Search the `OnClause`s, find the matching table identifiers, and we're set.

Aside: Oh hell, that's not necessarily going to work.
There's *two* ways to produce a `SqlExpr (Value (Key a))` - `entity ^. persistIdField` BUT ALSO `val (SomeKey a)`, which would render out to a mere `?`.
Well, I think that this approach may be dead in the water :(

Anyway, the algorithm is basically:

1. Find the rightmost `FromJoin` that has a `FromStart` as the immediate RHS (and does not already have an `OnClause`).
2. Find it's rightmost `FromStart` in the left branch.
3. Assign the `on` clause with matching table identifiers.
4. Set the `OnClause` for the `FromJoin` in step 1 to the `SqlExpr` that was rendered to produce that.
5. Descend into the LHS.
6. Go back to 1, terminate if there are no matches.

So. Hm. How can we handle the case where it's a `E.val someKey :: SqlExpr (Value (Key SomeEntity))`?
Let's verify how it renders:

```haskell
    it "renders ? for a val" $ do
      expr <- run $ EI.renderExpr (val (PersonKey 0) ==. val (PersonKey 1))
      expr `shouldBe` "? = ?"
```

This is an *abuse* of `ON`, IMO - this sort of query would look like:

```sql
SELECT *
FROM foo
INNER JOIN bar
ON bar.foo_id = 3
  -- where 3 is assumed to be an ID of a Foo that we care about
;
```

While this is syntactically correct SQL, it's really more appropriate to write this as:

```sql
SELECT *
FROM foo
INNER JOIN bar
ON bar.foo_id = foo.id
WHERE foo.id = 3
;
```

Another potentially weird set of JOINs are like this:

```sql
SELECT *
FROM foo
INNER JOIN bar
ON true
```

which is exactly equivalent to a cross join.

```sql
SELECT *
FROM foo
LEFT JOIN bar
ON false
```

which is going to produce nothin but NULLs for the `bar`s.
Nevertheless, these are definitely syntactically valid, so we should consider them.
For the most part, these correspond to "things that fail to parse into at least a pair of table identifiers."

We have 53 uses of `on` in the work codebase, and none use anything funky or weird - all are standard.
So I feel pretty confident that this algorithm will work well for most cases.
As a fallback, we can run the current algorithm again after performing the smarter matching, which ... should probably pick out the right things.
Or possibly fail, but I think this will be well worth the pain.

### DIFF
--- a/esqueleto.cabal
+++ b/esqueleto.cabal
@@ -32,6 +32,7 @@ library
       Database.Esqueleto.Internal.Language
       Database.Esqueleto.Internal.Sql
       Database.Esqueleto.Internal.Internal
+      Database.Esqueleto.Internal.ExprParser
       Database.Esqueleto.MySQL
       Database.Esqueleto.PostgreSQL
       Database.Esqueleto.PostgreSQL.JSON
@@ -45,9 +46,11 @@ library
   build-depends:
       base >=4.8 && <5.0
     , aeson >=1.0
+    , attoparsec >= 0.13 && < 0.14
     , blaze-html
     , bytestring
     , conduit >=1.3
+    , containers
     , monad-logger
     , persistent >=2.10.0 && <2.11
     , resourcet >=1.2
@@ -74,6 +77,7 @@ test-suite mysql
   ghc-options: -Wall
   build-depends:
       base >=4.8 && <5.0
+    , attoparsec
     , blaze-html
     , bytestring
     , conduit >=1.3
@@ -109,6 +113,7 @@ test-suite postgresql
   build-depends:
       base >=4.8 && <5.0
     , aeson
+    , attoparsec
     , blaze-html
     , bytestring
     , conduit >=1.3
@@ -143,6 +148,7 @@ test-suite sqlite
   ghc-options: -Wall
   build-depends:
       base >=4.8 && <5.0
+    , attoparsec
     , blaze-html
     , bytestring
     , conduit >=1.3
--- a/esqueleto.cabal
+++ b/esqueleto.cabal
@@ -31,12 +31,12 @@ library
       Database.Esqueleto
       Database.Esqueleto.Internal.Language
       Database.Esqueleto.Internal.Sql
+      Database.Esqueleto.Internal.Internal
       Database.Esqueleto.MySQL
       Database.Esqueleto.PostgreSQL
       Database.Esqueleto.PostgreSQL.JSON
       Database.Esqueleto.SQLite
   other-modules:
-      Database.Esqueleto.Internal.Internal
       Database.Esqueleto.Internal.PersistentImport
       Database.Esqueleto.PostgreSQL.JSON.Instances
       Paths_esqueleto
--- a/esqueleto.cabal
+++ b/esqueleto.cabal
@@ -86,6 +86,7 @@ test-suite mysql
     , exceptions
     , hspec
     , monad-logger
+    , mtl
     , mysql
     , mysql-simple
     , persistent >=2.8.0 && <2.11
@@ -122,6 +123,7 @@ test-suite postgresql
     , exceptions
     , hspec
     , monad-logger
+    , mtl
     , persistent >=2.10.0 && <2.11
     , persistent-postgresql >= 2.10.0 && <2.11
     , persistent-template
@@ -157,6 +159,7 @@ test-suite sqlite
     , exceptions
     , hspec
     , monad-logger
+    , mtl
     , persistent >=2.8.0 && <2.11
     , persistent-sqlite
     , persistent-template
--- a/esqueleto.cabal
+++ b/esqueleto.cabal
@@ -2,7 +2,6 @@ cabal-version: 1.12
 
 name:           esqueleto
 version:        3.2.0
-version:        3.1.2
 synopsis:       Type-safe EDSL for SQL queries on persistent backends.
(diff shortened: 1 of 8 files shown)


parsonsmatt:
Well, all my tests are passing, so that's cool!

Next steps:

- A *lot* more tests. I don't feel very confident that this is really doing exactly what I want in all cases.
- Clean it up
- Release!!!!!

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -105,71 +114,75 @@ fromJoin
...
 where_ :: SqlExpr (Value Bool) -> SqlQuery ()
 where_ expr = Q $ W.tell mempty { sdWhereClause = Where expr }
 
--- | @ON@ clause: restrict the a @JOIN@'s result.  The @ON@
--- clause will be applied to the /last/ @JOIN@ that does not
--- have an @ON@ clause yet.  If there are no @JOIN@s without
--- @ON@ clauses (either because you didn't do any @JOIN@, or
--- because all @JOIN@s already have their own @ON@ clauses), a
--- runtime exception 'OnClauseWithoutMatchingJoinException' is
--- thrown.  @ON@ clauses are optional when doing @JOIN@s.
```
parsonsmatt: yeet

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -105,71 +114,75 @@ fromJoin
...
 --   'on' (foo '^.' FooId '==.' bar '^.' BarFooId)
 --   ...
 -- @
 --
--- The order is /reversed/ in order to improve composability.
--- For example, consider @query1@ and @query2@ below:
+-- Old versions of esqueleto required that you provide the 'on' clauses in
+-- reverse order. This restriction has been lifted - you can now provide 'on'
+-- clauses in any order, and the SQL should work itself out. The above query is
+-- now totally equivalent to this:
```
parsonsmatt: :tada:

test/Common/Test.hs
```diff
@@ -446,7 +492,7 @@ testSelectFrom run = do
 
 testSelectJoin :: Run -> Spec
 testSelectJoin run = do
-  describe "select/JOIN" $ do
+  describe "select:JOIN" $ do
```
parsonsmatt: the slash makes it impossible to do a --test-arguments match on it

esqueleto.cabal
```diff
@@ -1,6 +1,7 @@
 cabal-version: 1.12
 
 name:           esqueleto
+version:        3.2.0
 version:        3.1.2
```
bitemyapp: Are there supposed to be two version lines here? What would that mean?

parsonsmatt: Bad merge conflict, thanks for catching!

Issue #150, opened 2019-10-10
*OuterJoin constructors should enforce nullability


parsonsmatt:
We're running into a few bugs at work now where someone did:

```haskell
select $
from $ \(a `LeftOuterJoin` b) -> do
where_ (a ^. AField ==. b ^. BField)
pure (a, b ^. BOtherField)
```

and we get `PersistMarshallException` trying to parse the `null` value out. These should be enforced to be a `Maybe`.

belevy:
Fixed in #172

PR #157, opened 2019-10-23
add associateJoin function

tippenein:
This function might want to live in esqueleto for avoiding N+1 queries? Just a rough example here along with (probably insufficient) code comment. It also adds a `containers` dep which might be undesirable.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,9 @@
+3.1.2
+========
+
+- @tippenein
+  - [#149](https://github.com/bitemyapp/esqueleto/pull/157): Added `associateJoin` query helpers.
+
 3.1.1
 ========
 
--- a/src/Database/Esqueleto/Internal/Internal.hs
+++ b/src/Database/Esqueleto/Internal/Internal.hs
@@ -50,7 +50,6 @@ import qualified Data.ByteString as B
 import qualified Data.Conduit as C
 import qualified Data.Conduit.List as CL
 import qualified Data.HashSet as HS
-import qualified Data.Map.Strict as Map
 import qualified Data.Text as T
 import qualified Data.Text.Lazy as TL
 import qualified Data.Text.Lazy.Builder as TLB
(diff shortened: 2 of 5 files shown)


parsonsmatt:
It looks like the lower bound on `containers` is too restrictive for what esqueleto supports. I'd guess that it's fine to relax that?

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -2884,3 +2885,27 @@ insertSelect = void . insertSelectCount
...
+-- | Avoid N+1 queries and join entities into a map structure
+-- @
+-- getFoosAndNestedBarsFromParent :: ParentId -> (Map (Key Foo) (Foo, [Entity Bar]))
+-- getFoosAndNestedBarsFromParent parentId = 'fmap' associateJoin $ 'select' $
+-- 'from' $ \\(foo `'LeftOuterJoin`` bar) -> do
+--   'on' (bar '^.' BarFooId '==.' foo '^.' FooId)
+--   'where_' (foo '^.' FooParentId '==.' 'val' parentId)
+--   'pure' (foo, bar)
+-- @
+
```
parsonsmatt: Please add a `@since` annotation with the relevant version.

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -2884,3 +2885,27 @@ insertSelect = void . insertSelectCount
 insertSelectCount :: (MonadIO m, PersistEntity a) =>
   SqlQuery (SqlExpr (Insertion a)) -> SqlWriteT m Int64
 insertSelectCount = rawEsqueleto INSERT_INTO . fmap EInsertFinal
+
+-- | Avoid N+1 queries and join entities into a map structure
+-- @
+-- getFoosAndNestedBarsFromParent :: ParentId -> (Map (Key Foo) (Foo, [Entity Bar]))
+-- getFoosAndNestedBarsFromParent parentId = 'fmap' associateJoin $ 'select' $
+-- 'from' $ \\(foo `'LeftOuterJoin`` bar) -> do
+--   'on' (bar '^.' BarFooId '==.' foo '^.' FooId)
```
parsonsmatt: The `bar` access here should use `?.` because a `LeftOuterJoin` implies nullability on the right hand side. That this isn't enforced is a bug (see #150). The type of this would then be `Map (Key Foo) (Foo, [Maybe (Entity Bar)])`.

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -2884,3 +2885,27 @@ insertSelect = void . insertSelectCount
...
+  => [(Entity e0, e1)]
+  -> Map.Map (Key e0) (e0, [e1])
+associateJoin = foldr f start
+  where
+    start = Map.empty
+    f (one, many) =
+      Map.insertWith
+        (\(oneOld, manyOld) (_, manyNew) -> (oneOld, manyNew ++ manyOld ))
+        (entityKey one)
+        (entityVal one, [many])
```
parsonsmatt: I think I'd define this directly in `Database.Esqueleto` and not one of the internal modules. It's easy to miss the reexports necessary.

changelog.md
```diff
@@ -1,3 +1,9 @@
+3.1.2
+========
+
+- @tippenein
+  - [#149](https://github.com/bitemyapp/esqueleto/pull/157): Added `associateJoin` query helpers.
```
parsonsmatt: :raised_hands:

PR #158, opened 2019-10-29
Add unsafe documentation

JoseD92:
Add documentation on how to use the unsafe functions and some of the pitfalls when using them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -256,7 +256,7 @@ type when you want to get results back from the database so better set types ear
 than late. This functions are also vulnerable to sql injection attacks so their use must be
 limited and with care, more on this later.
 
-The most common use of this functions is for calling RDBMS specific or custom functions,
+The most common use of these functions is for calling RDBMS specific or custom functions,
 for that end we use `unsafeSqlFunction`. For example, if we wish to consult the postgres
 `now` function we could so as follow:
 
--- a/README.md
+++ b/README.md
@@ -295,7 +295,7 @@ which generates this SQL:
 SELECT date_part('day', TIMESTAMP '2001-02-16 20:38:40')
 ```
 
-Using `unsafeSqlValue` we were able to also define the type of the value.
+Using `unsafeSqlValue` we were required to also define the type of the value.
 
 Another useful unsafe function is `unsafeSqlCastAs`, which allows us to cast any type
 to another within a query. For example, say we want to use our previews `dayPart` function
--- a/README.md
+++ b/README.md
@@ -246,7 +246,7 @@ In order to use these functions, you need to explicitly import their correspondi
 ### Unsafe functions, operators and values
 
 Not all RDBMSs specific functions and operators are expose on each RDBMS module, there is
-also the problem of calling user defined functions. To overcome this problem Esqueleto
+also the problem of calling user defined functions. To overcome this problem, Esqueleto
 exports a number of unsafe functions to call any function, operator or value. These functions
 can be found in Database.Esqueleto.Internal.Sql module.
 
--- a/README.md
+++ b/README.md
@@ -345,7 +345,7 @@ SELECT 'hi''; DROP TABLE foo; select ''bye'''
 ```
 
 And the printed value is `hi\'; DROP TABLE foo; select \'bye\'` and no table is dropped. This is good
-and makes the use of strings values save, unfortunately this is not the case when using unsafe functions.
+and makes the use of strings values safe. Unfortunately this is not the case when using unsafe functions.
 Let's see an example of defining a new evil `now` function:
 
 ```haskell
--- a/README.md
+++ b/README.md
@@ -243,6 +243,150 @@ There are many differences between SQL syntax and functions supported by differe
 
 In order to use these functions, you need to explicitly import their corresponding modules.
 
+### Unsafe functions, operators and values
+
+Not all RDBMSs specific functions and operators are expose on each RDBMS module, there is
+also the problem of calling user defined functions. To overcome this problem Esqueleto
+exports a number of unsafe functions to call any function, operator or value. This functions
+can be found on Database.Esqueleto.Internal.Sql module.
+
+Warning: the functions discuss on section should always be use with an explicit type signature
+as they are too general in order to maintain type safety, and you will need to indicate some
+type when you want to get results back from the database so better set types earlier rather 
+than late. This functions are also vulnerable to sql injection attacks so their use must be
+limited and with care, more on this later.
+
+The most common use of this functions is for calling RDBMS specific or custom functions,
+for that end we use `unsafeSqlFunction`. For example, if we wish to consult the postgres
+`now` function we could so as follow:
+
+```haskell
+postgresTime :: (MonadIO m, MonadLogger m) => SqlWriteT m UTCTime
+postgresTime = 
+  select (return now) >>= return . head
+  where
+    now :: SqlExpr (Value UTCTime) 
+    now = unsafeSqlFunction (fromText "now") ([] :: [SqlExpr (Value Int)])
+```
+
+which generates this SQL:
+
+```sql
+SELECT now()
+```
+
+With the `now` function we could now use the current time of the postgres RDBMS on any query.
+Do notice that `now` does not use any arguments, so we use an empty list cast to a valid
+`UnsafeSqlFunctionArgument` instance to represent no arguments. 
+
+Note: make () an instance of `UnsafeSqlFunctionArgument` for when no arguments are needed, this
+casting is quite dumb.
+
+We can also use `unsafeSqlFunction` for more complex functions with customs values using 
+`unsafeSqlValue` which turns any string into a sql value of whatever type we want, disclaimer:
+if you use it badly you will cause a runtime error. For example, say we want to try postgres'
+`date_part` function and get the day of a timestamp, we could use:
+
+```haskell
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -243,6 +243,150 @@ There are many differences between SQL syntax and functions supported by differe
...
+for that end we use `unsafeSqlFunction`. For example, if we wish to consult the postgres
+`now` function we could so as follow:
+
+```haskell
+postgresTime :: (MonadIO m, MonadLogger m) => SqlWriteT m UTCTime
+postgresTime = 
+  select (return now) >>= return . head
+  where
+    now :: SqlExpr (Value UTCTime) 
+    now = unsafeSqlFunction (fromText "now") ([] :: [SqlExpr (Value Int)])
```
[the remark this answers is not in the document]
parsonsmatt: Also, we don't need to use `fromText` because `OverloadedStrings` can handle that for us.

README.md
```diff
@@ -243,6 +243,150 @@ There are many differences between SQL syntax and functions supported by differe
...
+
+```haskell
+postgresTimestampDay :: (MonadIO m, MonadLogger m) => SqlWriteT m Int
+postgresTimestampDay = 
+  select (return $ dayPart date) >>= return . head
+  where
+    dayPart :: SqlExpr (Value UTCTime) -> SqlExpr (Value Int) 
+    dayPart s = unsafeSqlFunction (fromText "date_part") (unsafeSqlValue (fromText "\'day\'") :: SqlExpr (Value String) ,s)
+    date :: SqlExpr (Value UTCTime)
+    date = unsafeSqlValue . fromText $ "TIMESTAMP \'2001-02-16 20:38:40\'"
```
parsonsmatt: I think this would be a better example for `unsafeSqlCastAs` - this can also be written as:

```haskell
postgresTimestampDay :: (MonadIO m, MonadLogger m) => UTCTime -> SqlPersistT m Int
postgresTimestampDay time =
  head <$> select (return (dayPart time))
  where
    addTimestampCast :: SqlExpr (Value UTCTime) -> SqlExpr (Value UTCTime)
    addTimestampCast time = unsafeSqlCastAs "TIMESTAMP" time

    dayPart :: SqlExpr (Value UTCTime) -> SqlExpr (Value Int)
    dayPart t = unsafeSqlFunction "date_part" (val "day", addTimestampCast t)
```

Note that we don't need `unsafeSqlValue` for `day` because `val :: Text -> SqlExpr (Value Text)` works just fine here. But if we try to pass in the UTCTime without a timestamp, it gets confused:

```
λ> now <- getCurrentTime
λ> dexDB $ select $ pure (unsafeSqlFunction "date_part" (val "Day", val now))
[Debug#SQL] SELECT date_part(?, ?)
; [PersistText "Day",PersistUTCTime 2019-10-29 15:12:06.381663489 UTC]
[Error#dexter] An error occurred: [SqlError]: SqlError {sqlState = "42725", sqlExecStatus = FatalError, sqlErrorMsg = "function date_part(unknown, unknown) is not unique", sqlErrorDetail = "", sqlErrorHint = "Could not choose a best candidate function. You might need to add explicit type casts."}
*** Exception: SqlError {sqlState = "42725", sqlExecStatus = FatalError, sqlErrorMsg = "function date_part(unknown, unknown) is not unique", sqlErrorDetail = "", sqlErrorHint = "Could not choose a best candidate function. You might need to add explicit type casts."}

λ> dexDB $ select $ pure (unsafeSqlFunction "date_part" (val "Day", unsafeSqlCastAs "TIMESTAMP" (val now) :: SqlExpr (Value UTCTime)))
[Debug#SQL] SELECT date_part(?, CAST(? AS TIMESTAMP))
; [PersistText "Day",PersistUTCTime 2019-10-29 15:12:06.381663489 UTC]
[Value {unValue = 29.0}]
```

parsonsmatt: I think a better example may be MySQL global variables: https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html These are explicitly *not* functions syntactically, and it wouldn't work to use `val` either - you have to input the text directly.

README.md
```diff
@@ -243,6 +243,150 @@ There are many differences between SQL syntax and functions supported by differe
...
+```
+
+which generates this SQL:
+
+```sql
+SELECT date_part('day', TIMESTAMP '2001-02-16 20:38:40')
+```
+
+Using `unsafeSqlValue` we were able to also define the type of the value.
```
parsonsmatt:
```suggestion
Using `unsafeSqlValue` we were required to also define the type of the value.
```

README.md
```diff
@@ -243,6 +243,150 @@ There are many differences between SQL syntax and functions supported by differe
...
+```haskell
+postgresTimestampDay :: (MonadIO m, MonadLogger m) => SqlWriteT m Int
+postgresTimestampDay = 
+  liftIO getCurrentTime >>= \t -> select (return $ dayPart (toTIMESTAMP $ val t)) >>= return . head
+  where
+    dayPart :: SqlExpr (Value UTCTime) -> SqlExpr (Value Int) 
+    dayPart s = unsafeSqlFunction (fromText "date_part") (unsafeSqlValue (fromText "\'day\'") :: SqlExpr (Value String) ,s)
+    toTIMESTAMP :: SqlExpr (Value UTCTime) -> SqlExpr (Value UTCTime)
+    toTIMESTAMP = unsafeSqlCastAs "TIMESTAMP"
+```
```
parsonsmatt: Ahh, nice, you did that already! :smile:

README.md
```diff
@@ -243,6 +243,150 @@ There are many differences between SQL syntax and functions supported by differe
...
+Using `unsafeSqlValue` we were able to also define the type of the value.
+
+Another useful unsafe function is `unsafeSqlCastAs`, which allows us to cast any type
+to another within a query. For example, say we want to use our previews `dayPart` function
+on the current system time, we could:
+
+```haskell
+postgresTimestampDay :: (MonadIO m, MonadLogger m) => SqlWriteT m Int
+postgresTimestampDay = 
+  liftIO getCurrentTime >>= \t -> select (return $ dayPart (toTIMESTAMP $ val t)) >>= return . head
```
parsonsmatt: I would really prefer to avoid operators where they make the code less readable. `do` notation would read very nicely here, IMO.

README.md
```diff
@@ -243,6 +243,150 @@ There are many differences between SQL syntax and functions supported by differe
...
+  select (return $ val ("hi\'; DROP TABLE foo; select \'bye\'" :: String)) >>= liftIO . print
+```
+
+which generates this SQL(when using postgres):
+
+```sql
+SELECT 'hi''; DROP TABLE foo; select ''bye'''
+```
+
+And the printed value is `hi\'; DROP TABLE foo; select \'bye\'` and no table is drop. This is good
```
parsonsmatt:
```suggestion
And the printed value is `hi\'; DROP TABLE foo; select \'bye\'` and no table is dropped. This is good
```

README.md
```diff
@@ -243,6 +243,150 @@ There are many differences between SQL syntax and functions supported by differe
...
+and makes the use of strings values save, unfortunately this is not the case when using unsafe functions.
```
parsonsmatt:
```suggestion
and makes the use of strings values safe. Unfortunately this is not the case when using unsafe functions.
```

README.md
```diff
@@ -243,6 +243,150 @@ There are many differences between SQL syntax and functions supported by differe
...
+
+which generates this SQL:
+
+```sql
+SELECT date_part('day', TIMESTAMP '2001-02-16 20:38:40');DROP TABLE bar; select (16)
+```
+
+This will print 16 and also erase the `bar` table. The main take away of this examples is to
+never use any user or third party input inside an unsafe function without first parsing it or
+heavily sanitizing the input.
```
parsonsmatt: :+1:

README.md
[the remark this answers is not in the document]
JoseD92: oh well I was trying to be a bit more explicit with `fromText`

README.md
```diff
@@ -243,6 +243,150 @@ There are many differences between SQL syntax and functions supported by differe
...
+limited and with care, more on this later.
+
+The most common use of this functions is for calling RDBMS specific or custom functions,
+for that end we use `unsafeSqlFunction`. For example, if we wish to consult the postgres
+`now` function we could so as follow:
+
+```haskell
+postgresTime :: (MonadIO m, MonadLogger m) => SqlWriteT m UTCTime
+postgresTime = 
+  select (return now) >>= return . head
```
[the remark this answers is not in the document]
JoseD92: I am not sure, in this case we are guaranteed to have a result, and knowing when to use unsafe functions like `head` correctly is also important

README.md
JoseD92: yeah :) the other example with the `unsafeSqlValue` was to show that you could specify the type inside the string for `unsafeSqlValue` and this one to transform a value to what you need

README.md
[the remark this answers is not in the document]
parsonsmatt: We *are* guaranteed, but we might also be mistaken about that guarantee, or the underlying guarantee might change, or something in the surrounding code might change, and then we're debugging a much less helpful message. Annotating unsafe stuff with a message is IMO the proper way to go.

README.md
JoseD92: there is an example of `unsafeSqlCastAs` next to this one, I am closing this suggestion

PR #163, opened 2019-11-21
Allow unsafeSqlFunction to take up to 10 args

### LINKED ISSUE #162: Add UnsafeSqlFunctionArgument instances for bigger tuples?
At the moment, the `UnsafeSqlFunctionArgument` instances only support sql functions of up to 4 arguments. However, there are functions you might want to call which accept more than 4 arguments. For example: postgres' `make_interval`, which takes 7 (documented at https://www.postgresql.org/docs/10/functions-datetime.html). Would a PR adding instances for functions with up to 7 (or perhaps more) arguments be considered?

hdgarrood:
... without needing to nest tuples. Fixes #162. I've also added a bit more documentation to the `UnsafeSqlFunctionArgument` class, as the fact that you can nest tuples wasn't clear to me until I actually started working on this, and I thought it was mentioning in case anyone else does want to use `unsafeSqlFunction` with more than 10 arguments in the future.

After submitting your PR:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,9 @@
+3.2.3
+========
+
+- @hdgarrood
+  - [#163](https://github.com/bitemyapp/esqueleto/pull/163): Allow `unsafeSqlFunction` to take up to 10 arguments without needing to nest tuples.
+
 3.2.2
 ========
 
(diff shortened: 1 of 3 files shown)


parsonsmatt:
Thanks!

PR #166, opened 2019-12-12
Add several common SQL string functions

charukiewicz:
This PR adds several common SQL string functions we noticed were not implemented in the library:
- UPPER
- TRIM
- LTRIM
- RTRIM
- LENGTH
- LEFT
- RIGHT

------------

After submitting your PR:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,9 @@
+3.3.0
+========
+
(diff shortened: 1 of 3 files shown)


parsonsmatt:
This is great, thanks!

Issue #70, opened 2018-01-30
Selecting from sub query? (might not be an issue?)


Vlix:
I'm trying to use the window functions I've recently implemented in my own fork of Esqueleto, which at some point I'd want to try and make a pull request for Esqueleto proper, but I'm trying to do some custom queries with it, and I'm running into a limitation. (Or I'm dumb and don't realize this is totally doable in current Esqueleto)

I'm trying to construct this query:
```sql
select * from message
  where id in
    (select id
      from (select id,
              row_number() over (partition by chat_id order by time desc) as rn
              from message
            ) as x
      where x.rn <= 10
    );
```

I've managed to implement the following (which works, in Haskell as well as the resulting SQL query)
```sql
select id,
  row_number() over (partition by chat_id order by time desc) as rn
  from message;
```

But how would you select the id column from the resulting `SqlExpr (Value a)` and use the `x.rn <= 10` in the sub query around it?
`sub_select` will not allow the following `return`
```haskell
...
return (msg ^. MessageId, rn msg)
  where rn msg = row_number `over` (Just $ msg ^. MessageChatId)
                                 $ [desc $ msg ^. MessageTime]
```

Better yet, how do you select from a sub query?

bitemyapp:
Can you do it as a `sub_select`? I'd need to dig in with working code to say or do much further.

Vlix:
So this is what I've tried, but I get the resulting error shown below, so `sub_select` can't do this, or I'm using it incorrectly?

(This is the type signature of `over`, by the way)
```haskell
over :: SqlExpr (Value (WindowType a))
     -> Maybe (SqlExpr (Value b))
     -> [SqlExpr OrderBy]
     -> SqlExpr (Value a)
```

```haskell
getFromDB :: ReaderT SqlBackend IO [Entity Message]
getFromDB = select $ from $ \msg -> do
    where_ $ msg ^. MessageId `in_` subListQ
    return msg
  where subListQ = subList_select $ from $ \msg -> do
                      -- Something here that makes `msg`
                      -- come/be from `lastSubQ`?
                      return $ msg ^. MessageId
        lastSubQ = sub_select $ from $ \msg -> return (msg ^. MessageId, rn msg)
          where rn msg = row_number `over` (Just $ msg ^. MessageChatId)
                                         $ [desc $ msg ^. MessageTime]
```
```
.../DB.hs:107:33: error:
[...]

PR #168, opened 2019-12-22
From sub query

belevy:
Sorry to drop this on you like this. I tried to communicate on #70 

This is basically the same thing that jprider had in his but cleaned up. I decided to merge the aliases into values and entities so that the end user doesn't have to think about whether something is an alias or not. This turned out to help when i was trying to make sure that everything is implemented. I don't have great test coverage but I can add that. I mostly just want to know if this work is something that we want to get in to the master branch or if I barked up the completely wrong tree. I only have like 6 months of Haskell experience so you probably have some suggestions and I by no means am tied to this as the final implementation.

This currently supports only top level selects from queries, while this isn't particularly useful right now I implemented this to allow for support of SQL set operations. @parsonsmatt ran into issues with union at the beginning of the year related to this.

Joining on subqueries is probably a bit more complicated but I think can be based on this work. I was thinking about it and came up with the following idea. There would be top level functions for each of the join types i.e. `innerJoinQuery`, `leftOuterJoinQuery`. These would add the FromQuery to the join tree with the appropriate join kind. As far as on conditions I am not so sure but I feel like it should be doable. 

Once again sorry for dropping this on you from out of no where, I just felt that having working code was the best way to start a conversation.


belevy:
I implemented inner join in the manner I discussed above. I had to convert FromClause to be a Monoid to allow for the incremental addition of from clauses. In order to limit the complexity I just convert to a list of FromClause similar to the current version. As it is based on this PR but somewhat separate I feel like I shouldn't add it to this branch. The code is located at:
https://github.com/belevy/esqueleto/tree/monoidal-fromclause

Once again, thank you for taking the time to look at this and hopefully we can get this functionality into the project.

belevy:
I was able to implement what I think is a fairly good API for SQL Set Operations (UNION et al) based on the work done in this PR, so it is at the very least a good basis for implementing this kind of stuff. The API I came up with is

```
select $
fromSetOperation $
    (SelectQuery q1)
    `Union`
    (SelectQuery q2)
```

Of course I am not married to this and fully willing to rework it based on any differences for how sub queries work in the end.

You can view the code that backs this API at 
https://github.com/belevy/esqueleto/tree/union-prototype

parsonsmatt:
Thanks for all this work! I'll try and review it tomorrow.

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -531,15 +651,24 @@ val v = ERaw Never $ const ("?", [toPersistValue v])
...
-isNothing (ERaw p f)        = ERaw Parens $ first ((<> " IS NULL") . parensM p) . f
 isNothing (ECompositeKey f) = ERaw Parens $ flip (,) [] . (intersperseB " AND " . map (<> " IS NULL")) . f
+isNothing v = ERaw Parens $ first ((<> " IS NULL")) . g
+  where 
+    g =
+      case v of 
+        ERaw p f             -> first (parensM p) . f
+        EAliasedValue i _    -> aliasedValueIdentToRawSql i
+        EValueReference i i' -> valueReferenceToRawSql i i'
+        ECompositeKey _      -> undefined -- defined above
```
parsonsmatt: Can you restructure this code so that we don't have an unsafe pattern match here?

test/Common/Test.hs
```diff
@@ -841,6 +841,46 @@ testSelectJoin run = do
...
+        let q = from $ \( lord `InnerJoin` deed ) -> do
+                on $ lord ^. LordId ==. deed ^. DeedOwnerId
+                return (lord ^. LordId, deed ^. DeedId)
+
+        (ret :: [(Value (Key Lord), Value Int)]) <- select $ fromQuery q $ \(lordId,deedId) -> do
+                 groupBy (lordId)
+                 return (lordId, count deedId)
+
+        liftIO $ ret `shouldMatchList` [ (Value l3k, Value 7)
+                                       , (Value l1k, Value 3) ]
```
parsonsmatt: These test cases are great, but I think I'm a little confused on what this brings that is novel. Unless I'm mistaken, all of these can be written right now:

```haskell
-- groupBy
let 
  q = 
    from $ \(lord `InnerJoin` deed) -> do
    on $ lord ^. LordId ==. deed ^. DeedOwnerId
    return (lord ^. LordId, deed ^. DeedId)
select $ do
  (lordId, deedId) <- q
  groupBy lordId
  return (lordId, count deedId)
```

Can you write some tests or examples that demonstrate the utility that this approach brings?

belevy: You are correct, all the examples in the tests currently are just proving that things that were possible before are possible with selection from a subquery, I will try to figure out a case that couldn't be trivially rewritten. However, this is mainly just foundational work to support union/subquery joining.

parsonsmatt: That's cool! I think I want to see how union/subquery works out with this, or at least a path forward, before I merge it into the library. How much extra work do you imagine that might be?

belevy: https://github.com/belevy/esqueleto/tree/monoidal-fromclause has my work towards joining on subqueries. It follows the syntax i used above trying to be minimally invasive to the way things work right now. It only supports a single join and that has to be the first thing you do. I am working towards making that work. 

My work for unions is at https://github.com/belevy/esqueleto/tree/union-prototype. That I feel is more complete and is useful right now.

Additionally I was able to come up with a test case that is at least not trivial without fromQuery though it may be possible I am not sure. It involves aggregating along two dimensions while limiting the result set after the first aggregation. "How many lords with more than 3 deeds?"

belevy: fixed the single join only issue. But ordering still is incredibly important. All of the joins have to come before the on clauses start

belevy:
> I'm curious what the generated SQL is for your version and mine is, too.
Yours will error out as it will produce
```sql
SELECT COUNT(*) 
FROM `Lord`
INNER JOIN `Deed` ON `Lord`.`county` = `Deed`.`ownerId`
WHERE COUNT(*) > 3
GROUP BY `Lord`.`county`
```

Mine will produce
```sql
SELECT COUNT(*)
FROM (
   SELECT `Lord`.`county` as value, COUNT(*) as value2
   FROM `Lord`
   INNER JOIN `Deed` ON `Lord`.`county` = `Deed`.`ownerId`
   GROUP BY `Lord`.`county`) as subquery
WHERE `subquery`.`value2` > 3
```

I'm not sure how you would use a having to do this. At the very least using a subquery to do this is more straightforward in my opinion.

belevy:
So what are next steps, are we waiting on figuring out a satisfactory solution to subquery joins before moving forward with this PR?

parsonsmatt:
Yeah, I think that I want to see an API for `UNION` and subquery joins before incorporating this into the API for `esqueleto`.

belevy:
merged with #172

PR #172, opened 2020-01-28
New syntax for Joins (Subquery + Union/Intersect/...)

belevy:
New API properly enforces Maybe on outer joins and requires an on clause for all joins in their construction.

Made some changes to the API proposed over in https://github.com/belevy/esqueleto/pull/1 
Still requires several test cases to display the flexibility.

This will fit nicely with the changes made to the FromClause to support subqueries and unions. They will just require a new FromParts constructor. 

I am definitely up for renaming FromParts.

The core of this implementation is the GADT FromParts. This allows us to both model the Entity-ness of the Tables and the nullity caused by different kinds of outer joins. The top level fromParts (subject to renaming but from' had a name collision) function will take one of these FromParts GADTS and will pull out the SqlExprs wrapped inside (guaranteed to always be a SqlExpr since there is no way to construct a FromParts that isn't around SqlExprs) and add the from clause generated from the FromParts to the SideData. 

TL;DR This code can live alongside the existing implementation since at the end of the day it just adds a fully formed FromClause to the sdFromClauses list and adding the work done for subqueries should be "trivial (TM)"


belevy:
Latest commit renames a bunch of the constructors in the GADT away from the ' suffixed convention and adds a new ToFromParts typeclass that allows the use of the existing data types like \`InnerJoin\` and \`LeftOuterJoin\` in conjunction with FromParts constructors like Table.

I am considering adding a function
```haskell
a `on_` b = (,)
```

This could be a nice syntactic sugar instead of the tuple syntax. so 
```haskell
Table @Person `InnerJoin` (Table @BlogPosts `on_` \(p,bp) -> p ^. PersonId ==. bp ^. BlogPostAuthorId)
```
hopefully it could have a high enough fixity that it obviates the need for parentheses.

belevy:
Pending changes to the syntax introduced in this PR. https://github.com/belevy/esqueleto/pull/2 should add the ability to select and join for subqueries and unions

belevy:
rewriting a module in our production code base using this has exposed a couple of things.
1. join order is incredibly important from an ergonomics perspective, it would be better if there were some way to make the function for the on condition take an arbitrary sized tuple that can contain any of the values that have been introduced to this point (Is this even possible)
2. Nested joins are actually pretty good in this syntax but it did expose an existing annoyance with the library. using `?.` on a nullable field will make a SqlExpr (Maybe (Maybe Value)) this is not great.
3. the nested tuples are really much more annoying than i had hoped, i have been struggling to get the correct number of opening parens. Once again flattening would be ideal here though i am not even sure if that is possible

belevy:
With this final change to the syntax adding the `:&` type operator I cant think of any more issues with the syntax. Type errors for this stuff are still pretty bad and don't really tell you what is wrong but I feel like that can be added on top of this at a later point.

parsonsmatt:
Usability of type errors is really important to me with this project, so bad type errors is a bit of a blocker. I'll try to look at this soon and see if I can suggest any improvements for that.

belevy:
> 
> 
> Usability of type errors is really important to me with this project, so bad type errors is a bit of a blocker. I'll try to look at this soon and see if I can suggest any improvements for that.

Ok, I was trying to figure out how to use the TypeError stuff but its real confusing. I managed to convert all the functional dependencies in Experimental into associated type synonyms. That work is on https://github.com/belevy/esqueleto/tree/type-family-conversion
I don't know if that is helpful but the errors did seem slightly less bad with the type family than with FD.

bitemyapp:
@belevy 

>I don't know if that is helpful but the errors did seem slightly less bad with the type family than with FD.

That's more or less what I would expect.

Could you do me a favor and make repros of the bad type errors? Perhaps in a sub-dir called "bad-examples" that isn't part of the stack.yaml with its own stack.yaml that depends on the library above. It might be easier to get a type-stronaut to look at the situation if they can just press a button.

I haven't been paying attention closely so if this is already available, I apologize and thank you.

belevy:
Added tests like @bitemyapp asked for. Tests can be found at `tests/new-join-compiler-errors` and run with stack build. Looks like stack doesn't support compiling only one executable so all the compiler errors are in one exe `bad-errors`

src/Database/Esqueleto/Experimental.hs
```diff
@@ -299,107 +323,124 @@ from parts = do
...
+         , ToAlias g
+         , ToAlias h
+         ) => ToAlias (a,b,c,d,e,f,g,h) where
   toAlias x = to8 <$> (toAlias $ from8 x)
 
-
 -- more tedious tuple magic 
-class ToAliasReference a b | a -> b where
+class (ToAliasReferenceT a ~ b) => ToAliasReference a b where
+  type ToAliasReferenceT a
```
parsonsmatt: If this is the design, you can drop the second param on the class. `class ToAliasReference a; type ToAliasReferenceT a` has the same meaning.

belevy:
@parsonsmatt @bitemyapp I was able to get it to print a custom error message when there is a missing on statement. The other cases all seemed to have fairly straightforward if verbose error messages. That verbosity however seems impossible to dispel as it is caused by the "in equation..." stuff and so if you have a large query the output will be large. At the very top it will say something good. Having said that all, as the author of the new syntax its kinda hard for me to come up with a good error message for the case of the missing on clause. I have a fairly bare bones "missing on clause in <JoinKind>" message that I'm sure could be spruced up a little.

src/Database/Esqueleto/Experimental.hs
belevy: Oh whoops, that was from when I was still coming to grips with type families, I think I will move that to being a closed family like the others.

parsonsmatt:
I'll try to play with this tomorrow and see what I can get

charukiewicz:
@parsonsmatt I just sat down with @belevy today and we added a lot of Haddock documentation to `Database.Esqueleto.Experimental`. We tried to introduce the motivation for this module and give several side-by-side examples of the current syntax vs. the new syntax this module introduces. We plan on creating a few more examples later this week that include subqueries and some of the other more complex operations this module enables. Happy to hear your feedback if you want to build the Haddock and take a look.

belevy:
Sorry about the delay, we just finished off the documentation and it should be ready for others to test out now. I have been using the experimental module in my personal projects for a while now and I think the syntax is stable enough to go out to the world. Do you think this is ready to be merged into master and pushed to Hackage? I think publishing will reduce barriers to testing out the new syntax and help us get better feedback for further improvements.

parsonsmatt:
Finally getting around to playing with this!

So far it's very good. UX concerns are what I am mostly inspecting, and as I use type-holes a lot, I'm concerned about how well it's going to play with typed holes.

I started writing a join and used a typed hole:

image

Get a bunch of hole fits:

image

But it seems like a lot of these are definitely spurious. Using `id` or `Just` both get this:

image

The `undefined` is getting called in this stuff:

```haskell
instance {-# OVERLAPPABLE #-} ToFrom (InnerJoin a b) where
  toFrom = undefined
instance {-# OVERLAPPABLE #-} ToFrom (LeftOuterJoin a b) where
  toFrom = undefined
instance {-# OVERLAPPABLE #-} ToFrom (RightOuterJoin a b) where
  toFrom = undefined
instance {-# OVERLAPPABLE #-} ToFrom (FullOuterJoin a b) where
  toFrom = undefined
```

Commenting out those instances definitely causes a compile-time error with `Just`, and it makes the typed-hole not particularly useful:

image


My other favorite trick, using `()` as a type hole, gets a much better error message:

image

src/Database/Esqueleto/Experimental.hs
```diff
@@ -0,0 +1,845 @@
...
+--
+-- @
+-- select $
+-- from $ Table @Person
+-- \`InnerJoin\` Table @BlogPost
+-- \`on\` (\\(p :& bP) ->
+--         p ^. PersonId ==. bP ^. BlogPostAuthorId)
+-- @
+--
+on :: ToFrom a => a -> b -> (a, b)
```
parsonsmatt:
```suggestion
on :: ToFrom a => a -> (b -> SqlExpr (Value Bool)) -> (a, b -> SqlExpr (Value Bool))
```

I've found that locally this improves the type inference massively.

image


image

parsonsmatt:
Alright, I'm extremely pleased using this so far! It's stupid awesome. The docs are fantastic too but they appear to not be escaping the type applications appropriately.

charukiewicz:
@parsonsmatt Thanks for catching those missing escapes in the docs. I just made the fix in a PR from copy of the repo into Ben's copy of the repo (belevy/esqueleto#7). He should be on tomorrow and will merge it in.

Vlix:
Ok, bit late, but this is legitimately awesome! I think I asked for this a year (maybe more) ago, and can't remember who started trying some stuff out, but if it's you ( @belevy ) or not, either way, thank you very much for this PR. It's amazing!
(Haven't used it yet, but very probably will in the near future)

PR #177, opened 2020-03-20
Update (^.) to fix natural key handling

### LINKED ISSUE #176: DeconstructionError when using primary key that wraps a primary key on another table
Hi, I think there's a bug with using a primary key on a table, that is itself wrapping a primary key on another table. As an example, here ArticleMetadata's primary key is an `ArticleId`, which is the primary key of `Article`.

```
  Article
    title String
    frontcoverNumber Int
    Foreign Frontcover fkfrontcover frontcoverNumber
    deriving Eq Show
  ArticleMetadata
    articleId ArticleId
    Primary articleId
    deriving Show
```

In this case, an esqueleto query that does `(val articleMetadataId)` will error with `SqlBinOpCompositeErr DeconstructionError`.

I added a test to demonstrate this here:

https://github.com/MercuryTechnologies/esqueleto/compare/master...MaxGabriel:primaryForeignKeyBug?expand=1

I also added a `Text` field to `DeconstructionError` to get a little more data. The test fails with this:

```
  test/Common/Test.hs:760:5:
  1) Tests that are common to all backends, select:JOIN, allows using a primary key that is itself a key of another table
       uncaught exception: EsqueletoError
       SqlBinOpCompositeErr (DeconstructionError "(\"?\",[PersistInt64 16])")
```

This appears to be the code throwing, modified to show the bad value from deconstruct:

```
    deconstruct :: (TLB.Builder, [PersistValue]) -> ([TLB.Builder], [PersistValue])
    deconstruct ("?", [PersistList vals]) = (replicate (length vals) "?", vals)
    deconstruct (b', []) = (TLB.fromLazyText <$> TL.splitOn "," (TLB.toLazyText b'), [])
    deconstruct bad = throw (SqlBinOpCompositeErr (DeconstructionError (T.pack $ show bad)))
```

Questions I have:

1. Does this seem like a bug to you? I think this is valid use of persistent/esqueleto but wanted to just double check
2. Any ideas how to handle this case?
3. ~Any ideas on a workaround?~ (Edit: found: use `in_` and `valList` with a single value)
4. Related, I think it would be nice give a more detailed error when deconstruction fails. Would it be reasonable to add a string or text argument to `DeconstructionError` so if this happens in the future, it can be debugged more easily? It looks to be from an internal module so it seems fine to add it even though it's a "breaking" change. If so my branch adds this, and I could improve the error message to be more clear.

https://github.com/bitemyapp/esqueleto/issues/36 is possibly related?

belevy:
Update the handling of id fields in the projection operator. Since persistent treats all natural keys as composite keys we can not rely on hasCompositeKey. 

Instead use entityKeyFields to treat surrogate and single field natural keys the same. 

Fixes #176

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,9 @@
+3.3.2
+========
+
+- @belevy
+  - [#177](https://github.com/bitemyapp/esqueleto/pull/177) Fix natural key handling in (^.)
+
 3.3.1.1
 ========
 
--- a/esqueleto.cabal
+++ b/esqueleto.cabal
@@ -1,7 +1,7 @@
 cabal-version: 1.12
 
 name:           esqueleto
-version:        3.3.1.1
+version:        3.3.2
 synopsis:       Type-safe EDSL for SQL queries on persistent backends.
 description:    @esqueleto@ is a bare bones, type-safe EDSL for SQL queries that works with unmodified @persistent@ SQL backends.  Its language closely resembles SQL, so you don't have to learn new concepts, just new syntax, and it's fairly easy to predict the generated SQL and optimize it for your backend. Most kinds of errors committed when writing SQL are caught as compile-time errors---although it is possible to write type-checked @esqueleto@ queries that fail at runtime.
                 .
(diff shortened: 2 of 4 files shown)


tysonzero:
Just spent a while debugging and ended up with this exact fix. So yeah I would love to see this merged.

parsonsmatt:
@belevy mind writing a changelog entry? This will be a bugfix version bump.

MaxGabriel:
Thanks @belevy !

PR #178, opened 2020-03-22
Address Nullability and SQL operators

belevy:
This PR attempts to remove the major pain points of esqueleto nullability. I could not figure out a way to make the type inference play nicely with functions that accept either a type or its variant but that probably could be resolved with a Proxy type class similar to SqlString. However, this would remove the need to use joinV in most cases as the operators will already flatten the maybes using the SqlMaybe type class and also removes the need to use COALESCE before using boolean operators.

### DIFF
--- a/esqueleto.cabal
+++ b/esqueleto.cabal
@@ -33,6 +33,7 @@ library
       Database.Esqueleto.Internal.Sql
       Database.Esqueleto.Internal.Internal
       Database.Esqueleto.Internal.ExprParser
+      Database.Esqueleto.Internal.SqlTypeable
       Database.Esqueleto.MySQL
       Database.Esqueleto.PostgreSQL
       Database.Esqueleto.PostgreSQL.JSON
@@ -60,6 +61,7 @@ library
     , transformers >=0.2
     , unliftio
     , unordered-containers >=0.2
+    , vector
   if impl(ghc >=8.0)
     ghc-options: -Wall -Wno-redundant-constraints
   else
--- a/src/Database/Esqueleto.hs
+++ b/src/Database/Esqueleto.hs
@@ -53,7 +53,7 @@ module Database.Esqueleto
              , subList_select, valList, justList
              , in_, notIn, exists, notExists
              , set, (=.), (+=.), (-=.), (*=.), (/=.)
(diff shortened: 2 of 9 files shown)


tysonzero:
It seems like the `Nullable` type class could be replaced with:

```
instance SqlTypeable a => SqlTypeable (Maybe a) where
  type SqlType' (Maybe a) = SqlType' a
```

tysonzero:
Doing the following seems to have some benefits and minimal costs:

```
class PersistFieldSql t => SqlTypeable t where
  type SqlType' t :: SqlType
  type SqlNullable :: Bool
...
instance SqlTypeable Bool where
  type SqlType' Bool = 'SqlBool
  type SqlNullable = 'False
...
instance (SqlTypeable a, SqlNullable a ~ 'False) => SqlTypeable (Maybe a) where
  type SqlType' (Maybe a) = SqlType' a
  type SqlNullable = 'True
```

This would help address the nested `Maybe` problem where `Just Nothing` and `Nothing` are identical and blurred together.

It would also give more flexibility, by allowing non-Maybe types to be used for nullability as well if the developer desires.

tysonzero:
```
instance (PersistFieldSql (Maybe t), SqlTypeable t, SqlNullable t ~ 'False) => SqlTypeable (Maybe t) where
  type SqlType' (Maybe t) = SqlType' t
  type SqlNullable (Maybe t) = 'True
```

This is currently unsatisfiable. You need `persistent` to provide a `PersistFieldSql a => PersistFieldSql (Maybe a)` instance. Not sure why it doesn't already have one.

```
instance (PersistFieldSql (Key t)) => SqlTypeable (Key t) where
  type SqlType' (Key t) = 'SqlInt64
  type SqlNullable (Key t) = 'False
```

This is going to break for people like me who use something other than `Int64` for their primary keys (I use `UUID`).

It will probably be best to try and get some of this stuff merged directly into `persistent`.

belevy:
I don't really like how this path is evolving so I'm gonna close this PR. Work needs to be done on the persistent side to make a safe cast/coerce feasible. That said I still feel like a magic type family on maybe will make 90% of the headaches around maybe go away.

PR #180, opened 2020-03-30
Postgresql Date Truncation

### LINKED ISSUE #27: tried to use `date_trunc` in postgres, but it messes up UTC times
i wrote the following

````
truncateDate 
    :: E.SqlExpr (E.Value String)  -- ^ .e.g (val "day")
    -> E.SqlExpr (E.Value UTCTime) -- ^ input field
    -> E.SqlExpr (E.Value UTCTime) -- ^ truncated date
truncateDate datePart expr = unsafeSqlFunction "date_trunc" (datePart, expr)
````

but it seems to lose some kind of UTC magic.

if i use this, then in postgres  times like `2017-04-16 16:37:32.090688+10` become `2017-04-16 00:00:00+10` and when i bring them down to haskell and format them with, say, `formatTime defaultTimeLocale "%Y-%m-%d"` then i get "2017-04-15"

however, it i simply don't use my `truncateDate` function, then the exact same pipeline yield the "proper" date, "2017-04-16".

anyone have any idea what is happening here?

(this is in LTS-6.10, so it is esqueleto-2.4.3)

### PR DESCRIPTION
**Pull request by parsonsmatt**

After submitting your PR:

This PR includes a test for date truncation in postgresql. 

Fixes #27 

The observed behavior is somewhat weird on my machine locally:

```
  test/PostgreSQL/Test.hs:536:17:
  1) PostgreSQL specific tests, date_trunc, works
       Falsifiable (after 3 tests and 4 shrinks):
         [(0,0,0,0)]
       expected: 1999-12-31 07:00:00 UTC
        but got: 2000-01-01 00:00:00 UTC
```

The date is generated from this:

```haskell
    prop "works" $ \listOfDateParts -> run $ do
      let
        utcTimes =
          map
            (\(y, m, d, s) ->
              fromInteger s
                `addUTCTime`
                  UTCTime (fromGregorian (2000 + y) m d) 0
            )
          listOfDateParts
```

So it's weird to me that we get a 07:00:00 at all! We're inserting `UTCTime (fromGregorian 2000 0 0) 0`, which becomes `2000-01-01 00:00:00 UTC`. So why are we getting `1999-12-31 07:00:00 UTC` back from Postgres?

It would appear that the time zone shift is occurring with `date_trunc`, because the value round trips out of the database just fine.

May be relevant: https://www.postgresqltutorial.com/postgresql-timestamp/

### DIFF
--- a/test/Common/Test.hs
+++ b/test/Common/Test.hs
@@ -55,9 +55,13 @@ module Common.Test
     , Numbers (..)
     , OneUnique(..)
     , Unique(..)
+    , DateTruncTest(..)
+    , DateTruncTestId
+    , Key(..)
     ) where
 
 import Data.Either
+import Data.Time
 import Control.Monad (forM_, replicateM, replicateM_, void)
 import Control.Monad.Reader (ask)
 import Control.Monad.Catch (MonadCatch)
@@ -228,6 +232,10 @@ share [mkPersist sqlSettings, mkMigrate "migrateAll"] [persistUpperCase|
     name      String
     joinOther JoinOtherId
     joinOne   JoinOneId
+    deriving Eq Show
+
+  DateTruncTest
+    created   UTCTime
     deriving Eq Show
 |]
 
@@ -862,10 +870,10 @@ testSelectSubQuery run = do
     it "works" $ do
       run $ do
         _ <- insert' p1
-        let q = do 
+        let q = do
                 p <- Experimental.from $ Table @Person
                 return ( p ^. PersonName, p ^. PersonAge)
-        ret <- select $ Experimental.from $ SelectQuery q        
+        ret <- select $ Experimental.from $ SelectQuery q
         liftIO $ ret `shouldBe` [ (Value $ personName p1, Value $ personAge p1) ]
 
     it "lets you order by alias" $ do
@@ -873,8 +881,8 @@ testSelectSubQuery run = do
         _ <- insert' p1
         _ <- insert' p3
         let q = do
-                (name, age) <- 
-                  Experimental.from $ SubQuery $ do 
+                (name, age) <-
+                  Experimental.from $ SubQuery $ do
                       p <- Experimental.from $ Table @Person
                       return ( p ^. PersonName, p ^. PersonAge)
                 orderBy [ asc age ]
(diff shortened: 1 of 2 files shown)


parsonsmatt:
Based on the discussion [here](https://dba.stackexchange.com/questions/59006/what-is-a-valid-use-case-for-using-timestamp-without-time-zone), it seems that we are observing some slightly odd behavior on parsing `UTCTime` in `persistent`.

```
[local] matt@matt=# select (timestamp with time zone '2000-01-01 00:00:00 UTC');     
     timestamptz       
------------------------
 1999-12-31 17:00:00-07
(1 row)
```

OK. Cool. So `postgres` is rendering the time in *my local time zone*. Let's `date_trunc` it.

```
[local] matt@matt=# select date_trunc('day', timestamp with time zone '2000-01-01 00:00:00 UTC');
       date_trunc       
------------------------
 1999-12-31 00:00:00-07
(1 row)

Time: 5.520 ms
```

OK so this just blows away the `H:M:S` part of it, but we still have the time zone recorded. So then, when we go to parse this in `persistent`, we pull out `UTCTime (fromGregorian 1999 12 31) 0`, and then we see the time zone `07` and we do `addUTCTime (7 * 60 * 60)` to get the answer in the original value.

Naturally, this is going to depend on your local time zone settings, which is quite unfortunate.

parsonsmatt:
Alternatively, one could write `at time zone 'utc' date`. See this behavior difference:

```
[local] matt@matt=# select (timestamp with time zone '2000-01-01 00:00:00 UTC') at time zone 'utc' date;
        date         
---------------------
 2000-01-01 00:00:00
(1 row)

Time: 0.467 ms
[local] matt@matt=# select (timestamp with time zone '2000-01-01 00:00:00 UTC') date;
          date          
------------------------
 1999-12-31 17:00:00-07
(1 row)

[local] matt@matt=# select ((timestamp with time zone '2000-01-01 00:00:00 UTC') at time zone 'utc') :: date;
  timezone  
------------
 2000-01-01
(1 row)
````

:|

parsonsmatt:
well, uh, Travis isn't showing up, but [this PR is building](https://travis-ci.org/github/bitemyapp/esqueleto/builds/668838315?utm_medium=notification&utm_source=github_status), so I'm going to go ahead and merge this in.

PR #186, opened 2020-05-26
Remove outdated note about ON clause ordering

mitchellvitez:
It looks like #156 fixed the issue that this part of the README is still referring to


parsonsmatt:
great find thanks!

PR #190, opened 2020-06-20
Test and further document ToBaseId

MaxGabriel:
My coworker Lev was adding this typeclass to our codebase and we hadn't used it before. I added a little more documentation that I think would help clarify things, particularly what the point of the witness function was. More importantly I added a test for this typeclass.

- [ ] Bumped the version number -- Note: I saw there were other unreleased changes, so I just added an "unreleased" section to the changelog with my changes. I can bump the version number if you'd like
- [x] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html) (No new APIs)
- [ ] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock (No new APIs)

After submitting your PR:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,8 @@
+Unreleased
+========
+- @maxgabriel
+  - [#190](https://github.com/bitemyapp/esqueleto/pull/190) Further document and test `ToBaseId`
+
 3.3.3.1
 ========
 - @belevy
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,4 @@
-Unreleased
+3.3.3.2
 ========
 - @maxgabriel
   - [#190](https://github.com/bitemyapp/esqueleto/pull/190) Further document and test `ToBaseId`
--- a/src/Database/Esqueleto/Internal/Internal.hs
+++ b/src/Database/Esqueleto/Internal/Internal.hs
@@ -996,7 +996,7 @@ case_ = unsafeSqlCase
 -- | Convert an entity's key into another entity's.
 --
 -- This function is to be used when you change an entity's @Id@ to be
--- that of another entity.  For example:
+-- that of another entity. For example:
 --
 -- @
 -- Bar
@@ -1007,12 +1007,13 @@ case_ = unsafeSqlCase
 --   Primary bar
 -- @
 --
--- For this example, declare:
+-- In this example, Bar is said to be the BaseEnt(ity), and Foo the child.
+-- To model this in Esqueleto, declare:
 --
 -- @
 -- instance ToBaseId Foo where
 --   type BaseEnt Foo = Bar
---   toBaseIdWitness = FooKey
+--   toBaseIdWitness barId = FooKey barId
 -- @
 --
 -- Now you're able to write queries such as:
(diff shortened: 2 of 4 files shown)


MaxGabriel:
image

changelog.md
```diff
@@ -1,3 +1,8 @@
+Unreleased
+========
+- @maxgabriel
+  - [#190](https://github.com/bitemyapp/esqueleto/pull/190) Further dcument and test `ToBaseId`
```
MaxGabriel, line 4:
```suggestion
  - [#190](https://github.com/bitemyapp/esqueleto/pull/190) Further document and test `ToBaseId`
```

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1007,12 +1007,13 @@ case_ = unsafeSqlCase
 --   Primary bar
 -- @
 --
--- For this example, declare:
+-- In this example, Bar is said to be the BaseEnt(ity), and Foo the child.
+-- To model this in Esqueleto, declare:
```
MaxGabriel, line 1011: Explicitly clarifying what is supposed to be the Base was helpful to me. It's kind of confusing b/c the code reads as `BaseEnt Foo`, which vaguely looks like Foo is the BaseEnt

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1007,12 +1007,13 @@ case_ = unsafeSqlCase
...
 --
 -- @
 -- instance ToBaseId Foo where
 --   type BaseEnt Foo = Bar
---   toBaseIdWitness = FooKey
+--   toBaseIdWitness barId = FooKey barId
```
MaxGabriel, line 1016: Lev and I thought removing the eta reduction makes it more clear what this function is doing

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1371,7 +1372,11 @@ instance SqlString a => SqlString (Maybe a) where
 -- | Class that enables one to use 'toBaseId' to convert an entity's
 -- key on a query into another (cf. 'toBaseId').
 class ToBaseId ent where
+  -- | e.g. @type BaseEnt MyBase = MyChild@
   type BaseEnt ent :: *
+  -- | Convert from the key of the BaseEnt(ity) to the key of the child entity.
+  -- This function is not actually called, but that it typechecks proves this operation is safe.
+  -- You may be able to succinctly implement this with 'Data.Coerce.coerce'
```
MaxGabriel, line 1379: I think coerce is both a nice implementation tip, but also gives good intuition for what this function is about

MaxGabriel: Hm, after thinking about it, it's probably a mistake to recommend `coerce`. Since:

```
ghci> :t coerce (UserKey UUID.nil) :: OrganizationId
coerce (UserKey UUID.nil) :: OrganizationId :: OrganizationId
```

is just fine, but is actually incorrect

parsonsmatt: Yeah, `coerce` is a little unsafe, in that it *does* work but it relies on the user guaranteeing that the two underlying IDs are indeed the same.

Leaving it in there is probably fine but we should leave a warning to the edge case.

MaxGabriel:
Ok cool, I removed coerce b/c it seems like bad advice to give people. I also rebased with master which resolved a conflict.

I saw that 3.3.1 was released from the last PR—would you like this to be 3.3.1.1, or just mark this as "unreleased" in the changelog

parsonsmatt:
I'm happy releasing this as 3.3.3.2

MaxGabriel:
Ok bumped to 3.3.3.2

PR #191, opened 2020-07-06
Bugfix rollup: Fix issue with extra characters in generated SQL; Fix ToAliasReference for already referenced values; Fix Alias/Reference for Maybe Entity

belevy:
`in_` was generating invalid mysql statements by double wrapping value lists. This existed before 3.3.3.1. Also fixed an issue with the generated value names in subqueries getting "`"s erroneously added for each nesting

After submitting your PR:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,13 +1,10 @@
-3.3.3.4
+3.3.3.3
 =======
 - @belevy
   - [#191](https://github.com/bitemyapp/esqueleto/pull/191) - Bugfix rollup: 
     Fix issue with extra characters in generated SQL;
     Fix ToAliasReference for already referenced values; 
     Fix Alias/Reference for Maybe Entity 
-
-3.3.3.3
-=======
 - @maxgabriel
   - [#203](https://github.com/bitemyapp/esqueleto/pull/203) Document `isNothing`
 - @sestrella
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,8 @@
+3.3.3.3
+========
+- @belevy
+  - [#191](https://github.com/bitemyapp/esqueleto/pull/191) - Fix issue with extra characters in generated SQL
+
 3.3.3.2
 ========
 - @maxgabriel
--- a/changelog.md
+++ b/changelog.md
@@ -1,7 +1,10 @@
 3.3.3.3
 ========
 - @belevy
-  - [#191](https://github.com/bitemyapp/esqueleto/pull/191) - Fix issue with extra characters in generated SQL
+  - [#191](https://github.com/bitemyapp/esqueleto/pull/191) - Bugfix rollup: 
+    Fix issue with extra characters in generated SQL; 
+    Fix ToAliasReference for already referenced values; 
+    Fix Alias/Reference for Maybe Entity 
 
 3.3.3.2
 ========
--- a/esqueleto.cabal
+++ b/esqueleto.cabal
@@ -1,7 +1,7 @@
 cabal-version: 1.12
 
 name:           esqueleto
-version:        3.3.3.2
+version:        3.3.3.3
 synopsis:       Type-safe EDSL for SQL queries on persistent backends.
 description:    @esqueleto@ is a bare bones, type-safe EDSL for SQL queries that works with unmodified @persistent@ SQL backends.  Its language closely resembles SQL, so you don't have to learn new concepts, just new syntax, and it's fairly easy to predict the generated SQL and optimize it for your backend. Most kinds of errors committed when writing SQL are caught as compile-time errors---although it is possible to write type-checked @esqueleto@ queries that fail at runtime.
                 .
--- a/esqueleto.cabal
+++ b/esqueleto.cabal
@@ -1,7 +1,7 @@
 cabal-version: 1.12
 
 name:           esqueleto
-version:        3.3.3.4
+version:        3.3.3.3
 synopsis:       Type-safe EDSL for SQL queries on persistent backends.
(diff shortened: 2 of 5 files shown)


belevy:
@parsonsmatt the code that triggered this bug was 
```
posts ^. FeedPostPermissionId `in_` valList [6..8]
```
unsafeSqlFunction isn't wrapping in parens at all but unsafeSqlBinOp was. Not sure if we should just Never add parens or if this safer check is better. Either way, bumped and were gonna be relying on this commit at work.

belevy:
Found another bug while working on this query. Was doing a SubQuery join on a Union and got errors where the union alias (`u`) was being referenced rather than the subquery (`q`) of the union. This led to mysql throwing an error `u`.`v_id` is not a field. Figured since this PR was still open, better not to create extra versions.

src/Database/Esqueleto/Experimental.hs
```diff
@@ -767,6 +771,7 @@ instance ( ToAlias a
 type family ToAliasReferenceT a where
   ToAliasReferenceT (SqlExpr (Value a)) = SqlExpr (Value a)
   ToAliasReferenceT (SqlExpr (Entity a)) = SqlExpr (Entity a)
+  ToAliasReferenceT (SqlExpr (Maybe (Entity a))) = SqlExpr (Maybe (Entity a))
```
googleson78, line 774: This is a very unrelated question, but why do both `ToAliasT` and `ToAliasReferenceT` exist? They seem to do the same thing (?)

belevy: They might be able to be merged but it's not clear to me that they will always be the same. They initially started as associated type families in their respective classes. Semantically they are different though, an alias is something like "value as alias" where an alias reference is like "q.v".

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -2888,11 +2894,15 @@ valueReferenceToRawSql sourceIdent columnIdentF info =
...
+  I (baseIdent <> "_" <> (removeTicksFromColumnName $ builderToText $ fromDBName info $ fieldDB field))
 
 aliasedColumnName :: Ident -> IdentInfo -> T.Text -> TLB.Builder 
 aliasedColumnName (I baseIdent) info columnName = 
-  useIdent info (I (baseIdent <> "_" <> columnName))
+  useIdent info (I (baseIdent <> "_" <> (removeTicksFromColumnName columnName)))
+
+removeTicksFromColumnName :: T.Text -> T.Text
+removeTicksFromColumnName columnName = 
+    Maybe.fromMaybe columnName $ T.stripPrefix "`" =<< T.stripSuffix "`" columnName
```
parsonsmatt, line 2905: The ticks are a MySQL thing. Are we sure that this isn't potentially a problem in Postgres too?

It may be more resilient to get the quote character from the SqlBackend if possible

belevy: Right so this is probably better solved by changing `Ident` to be `Ident DBName` and just deferring the call to `fromDBName` until the last possible moment. I will probably back out this specific part of the PR since everything should still work with out this and see about changing `Ident`

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -2250,8 +2250,15 @@ unsafeSqlFunctionParens :: UnsafeSqlFunctionArgument a =>
...
-          uncommas' $ map (\(ERaw p f) -> first (parensM p) (f info)) $ toArgList arg
+    let
+      valueToFunctionArgParens v =
+        case v of
+          ERaw p f             -> first (parensM p) (f info)
+          EAliasedValue i _    -> aliasedValueIdentToRawSql i info
+          EValueReference i i' -> valueReferenceToRawSql i i' info
+          ECompositeKey _      -> throw (CompositeKeyErr SqlFunctionError)
+      (argsTLB, argsVals) =
+          uncommas' $ map valueToFunctionArgParens $ toArgList arg
```
belevy, line 2261: fix #201

belevy:
@parsonsmatt I fixed #201 while while resolving the conflicts. Seemed better to fix it in one PR and its not much more code. See the commented code for the fix. Additionally removed the other ERaw lambda that was left but it uses the same code that unsafeSqlFunction uses

changelog.md
```diff
@@ -1,5 +1,13 @@
+3.3.3.4
```
parsonsmatt, line 1: We haven't released 3.3.3.3 yet, so let's fold these changes back in :)

esqueleto.cabal
```diff
@@ -1,7 +1,7 @@
 cabal-version: 1.12
 
 name:           esqueleto
-version:        3.3.3.3
+version:        3.3.3.4
```
```suggestion
version:        3.3.3.3
```

changelog.md
```suggestion
3.3.3.3
```

changelog.md
```diff
@@ -1,5 +1,13 @@
...
+=======
+- @belevy
+  - [#191](https://github.com/bitemyapp/esqueleto/pull/191) - Bugfix rollup: 
+    Fix issue with extra characters in generated SQL;
+    Fix ToAliasReference for already referenced values; 
+    Fix Alias/Reference for Maybe Entity 
+
 3.3.3.3
```
parsonsmatt, line 9:
```suggestion
```

changelog.md
```diff
@@ -1,5 +1,13 @@
...
-========
+=======
```
parsonsmatt, line 10:
```suggestion
```

Issue #194, opened 2020-07-10
Experimental: Union of queries with limit doesn't render properly


googleson78:
In mysql at least, when the queries being `UNION`'d have `LIMIT`s in them, you need extra parens around each subquery

belevy:
This is not standard sql behaviour. MySQL is the only engine I know that allows parentheses and limits/orderby. I do wonder if maybe we could do a check to see if limit and orderby are used in the queries being unioned and optionally wrap them

googleson78:
Could you elaborate on the first part? I'm an SQL newbie, is there some issue with just putting extra parens around `x` and `y` in this expression: 
```sql
x `UNION` y
```

It seems to me like redundant parens shouldn't invalidate the query (?)

wrt. the check, I remember spotting some `NeedParens` datatype and mechanism in `SqlExpr` for this purpose, perhaps it could be helpful here? (although `Experimental.from` directly generates `SqlQuery`s)

(and thanks for making `Experimental` and responding so fast! the `JOIN`s seem great, and I would've had to write raw sql for the union otherwise!)

belevy:
Yah, this isn't the only case of extra parentheses causing the sql engine to choke. For example `value in ((1,2,3))` is a sql error. MySQL is fine with parens in set operations but at least sqlite isn't and I'm fairly sure that postgres also doesn't like it.

googleson78:
In this case the issue is much worse, I think. I'm pretty sure at the moment of rendering `esqueleto` isn't aware of the backend being used. I'm not sure what the solution here is at all :/ 
* some way to plug this info (or "information on how to render unions") somewhere (?)
* creating some kind of separate version of `Experimental.from` for mysql in the mysql specific module

@parsonsmatt any clues?

parsonsmatt:
Supporting MySQL is *the worst* :joy: 

We can know via [`connRDBMS`](https://www.stackage.org/haddock/lts-16.4/persistent-2.10.5.2/Database-Persist-Sql.html#t:SqlBackend), and I think the `SqlBackend` is carried around in the `SqlQuery` monad.

PR #195, opened 2020-07-10
Add new SetOperation constructor for parenthesized query

belevy:
This is a potential fix for #194. I am not sure that I like the extra SelectQueryP constructor but I don't want to muddy up the api for others and cant really think of a nicer way to convey the information. 

I guess theoretically aliasQuery could use a different data type that is exactly the same but with SelectQuery taking a NeedParens. 
I don't know if that is better or worse, do we want to expose the ability to add parentheses to the user directly?

I am certainly not tied to this, in fact it seems super hacky but so is mysql. I think this is probably better than depending on connRDBMS == "mysql".

### DIFF
--- a/src/Database/Esqueleto/Experimental.hs
+++ b/src/Database/Esqueleto/Experimental.hs
@@ -68,6 +68,7 @@ import Database.Esqueleto.Internal.Internal
           , to3, to4, to5, to6, to7, to8
           , from3, from4, from5, from6, from7, from8
           , veryUnsafeCoerceSqlExprValue
+          , parens
           )
 import GHC.TypeLits
 
@@ -379,7 +380,8 @@ data SqlSetOperation a =
   | UnionAll (SqlSetOperation a) (SqlSetOperation a)
   | Except (SqlSetOperation a) (SqlSetOperation a)
   | Intersect (SqlSetOperation a) (SqlSetOperation a)
-  | SelectQuery (SqlQuery a)
+  | SelectQuery  (SqlQuery a)
+  | SelectQueryP (SqlQuery a)
 
 -- | Data type that represents the syntax of a 'JOIN' tree. In practice,
 -- only the @Table@ constructor is used directly when writing queries. For example,
@@ -463,6 +465,7 @@ instance {-# OVERLAPPABLE #-} ToFrom (FullOuterJoin a b) where
 instance (SqlSelect a' r,SqlSelect a'' r', ToAlias a, a' ~ ToAliasT a, ToAliasReference a', ToAliasReferenceT a' ~ a'')  => ToFrom (SqlSetOperation a) where
   -- If someone uses just a plain SelectQuery it should behave like a normal subquery
   toFrom (SelectQuery q) = SubQuery q
+  toFrom (SelectQueryP q) = SubQuery q
   -- Otherwise use the SqlSetOperation
   toFrom q = SqlSetOperation q
 
@@ -622,7 +625,18 @@ from parts = do
                   prevState <- Q $ lift S.get
                   aliasedRet <- toAlias ret
                   Q $ lift $ S.put prevState
-                  pure (SelectQuery $ Q $ W.WriterT $ pure (aliasedRet, sideData), aliasedRet)
+                  let selectConstructor = if    (sdLimitClause sideData) /= mempty 
+                                             || length (sdOrderByClause sideData) > 0 then
+                                            SelectQueryP
+                                          else 
+                                            SelectQuery
+                  pure (selectConstructor $ Q $ W.WriterT $ pure (aliasedRet, sideData), aliasedRet)
+                SelectQueryP q -> do
+                  (ret, sideData) <- Q $ W.censor (\_ -> mempty) $ W.listen $ unQ q
+                  prevState <- Q $ lift S.get
+                  aliasedRet <- toAlias ret
+                  Q $ lift $ S.put prevState
+                  pure (SelectQueryP $ Q $ W.WriterT $ pure (aliasedRet, sideData), aliasedRet)
                 Union     o1 o2 -> do
                   (o1', ret) <- aliasQueries o1
                   (o2', _  ) <- aliasQueries o2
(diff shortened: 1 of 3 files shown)


googleson78:
If you have the time+will, could you explain how you avoid the "adding parens when on sqlite" problem here? I don't get it

belevy:
So there is already a runtime error when using limit or order by. Since those aren't supported in the subqueries of a union according to the sql standard the code assumes the user is using MySQL and wraps them in parens. This is still a runtime error of course but I think it falls within the realm of reasonability. @parsonsmatt may have different views though, that is why I left this out of the bugfix roll-up pending approval

googleson78:
Funnily this non-standard thing works in MySQL (union of limits), but what is "the standard way" (I think? I didn't manage to find (it in) the actual sql standard):
```sql
select * from (select * from A limit 3)
union
select * from (select * from A limit 3)
```
doesn't work in mysql, because you can't have limit in subqueries.. :facepalm:

Thanks!

googleson78:
Another "how did you do that?":
Always adding parens causes an error if you have more than one union - mysql doesn't like this:
```sql
(select * from A limit 3) union all ((select * from A limit 3) union all (select * from A limit 3));
```
How did you avoid it here? esqueleto's code generation is still a mystery to me..

belevy:
> How did you avoid it here? esqueleto's code generation is still a mystery to me..

So the SetOperations take place in a two phase process. First the tree is traversed by the aliasQuery function, this change will determine if the SelectQuery leaves use either a limit or orderBy by examining the `SideData`(the core data structure of the query builder, Q is a WriterT SideData monad). Basically, we run the writer for the subquery and create a "Fake" query that just has the resulting sideData. This new fake query will be added as FromQuery in the FromClause section of the SideData. This changeset adds a new SelectQueryP constructor that will wrap whatever the results of the subquery are in parens. When the time comes to generate the string we traverse the tree again with recursive calls to operationsToSql by way of doSetOperation. The only thing that is getting wrapped are the SelectQueryP leaves as you can see on line 661. Hope you can parse the wall of text.

belevy:
Upon further investigation I now believe this to be the simplest solution. The SqlBackend is not available to the SqlQuery as far as I can tell. The only thing I am worried about with this implementation is that it is encouraging non standard behaviour in a common module. There is some lack of type safety here already since we don't really contextualize queries to determine which operations are valid(the limit example being just one case), I'm not even sure how we would do that in the current implementation and certainly not sure how we would handle this on a per engine basis without a radical overhaul of how queries work. 

@parsonsmatt do you have any strong feelings about this "trust the programmer" solution? Do you have any idea about how to make a more useful warning/error?

src/Database/Esqueleto/Experimental.hs
```diff
@@ -379,7 +380,8 @@ data SqlSetOperation a =
   | UnionAll (SqlSetOperation a) (SqlSetOperation a)
   | Except (SqlSetOperation a) (SqlSetOperation a)
   | Intersect (SqlSetOperation a) (SqlSetOperation a)
-  | SelectQuery (SqlQuery a)
+  | SelectQuery  (SqlQuery a)
+  | SelectQueryP (SqlQuery a)
```
parsonsmatt, line 384: Does it make sense to use something like `SelectQuery NeedsParens (SqlQuery a)` instead of this?

test/MySQL/Test.hs
```diff
@@ -162,6 +164,31 @@ testMysqlTextFunctions = do
...
+        let q2 = do
+              foo <- Experimental.from $ Table @Foo
+              where_ $ foo ^. FooName >. val 3
+              orderBy [asc $ foo ^. FooName]
+              limit 2
+              pure $ foo ^. FooName
+
+
+        ret <- select $ Experimental.from $ SelectQuery q1 `Union` SelectQuery q2
+        liftIO $ ret `shouldMatchList` [Value 1, Value 2, Value 4, Value 5]
```
parsonsmatt, line 190: thanks for tests :raised_hands:

parsonsmatt:
@belevy hmmmm, maybe it makes sense to include the `SqlBackend` into the `State` part of `SqlQuery`. Or, to be more pure, we could make it into a `ReaderT SqlBackend (State IdentState a)`. We have a *lot* of functions from `IdentInfo ~ (SqlBackend, IdentState)`, and it may simplify the codebase to carry that along.

src/Database/Esqueleto/Experimental.hs
belevy: i think it would require breaking the public API since SelectQuery is used directly. Maybe PatternSynonyms could fix that?

parsonsmatt: Oh shoot! That's nasty. 

Actually, since this is adding a constructor to a public type, that's a breaking change too... If we'd only exposed named functions then we could get away with adding constructors for internal use.

Maybe let's add a comment that users should only be *constructing* with these, and not pattern matching? And additional constructors may be added without a major version bump?

It's an `Experimental` module so I'm not too particular about breaking changes.

belevy: If we're gonna break the API then I'd rather break it properly and hide the constructors completely and replace them with smart constructors.

parsonsmatt: Same. 

Hm. Let's put deprecation notices on the constructors, so people get warns with "how to fix" and not breaks. Then we can hide them in another release and remove the deprecation notices.

Alternatively, we move the constructors to an `Internal` module, make pattern synonyms, warn on the pattern synonym, and point to the named functions. Then a future release can delete the pattern synonyms.

belevy: If we go the pattern synonyms route there isn't really a reason to deprecate the API. Not like there are going to be new set operations. It's just gonna be implementation details that potentially change.

parsonsmatt: Good point

parsonsmatt:
also yeah let's get this out with 3.3.3.3

belevy:
Ok so I unified SelectQuery and SelectQueryP and exposed a pattern synonym. This seemed like the least breaking change possible. Do you want this to go out with the assumption that the programmer knows what they are doing or does this need to looks at the SqlBackend to ensure you are using MySQL?
It is unfortunate that we still have a gap in the type system related to calling limit inside of a union but short of handling things like beam does and tracking what is legal with which backends I don't see a way around it. I feel like it would be different if sqlite or postgres supported limit inside of unions without parentheses but as it is the only backend where this is legal is MySQL

parsonsmatt:
`esqueleto` has already bitten the bullet of "not all type-checked queries are safe." I'd prefer to have as few of those as possible, but what else are we gonna do - throw a Haskell runtime error instead of a SQL runtime error?

PR #200, opened 2020-08-02
Removes irrelevant copy pasta from some haddocks

timmytofu:


bitemyapp:
I kinda like defensive editing in documentation like this but YMMV. I'm good with whatever @parsonsmatt thinks is best.

parsonsmatt:
These docs are definitely erroneous, from copy pasting the definition of `renderQuery` which accepts the MODE parameter.

PR #208, opened 2020-09-24
update nightly resolver

parsonsmatt:
a mere CI update


Vlix:
Wouldn't `resolver: nightly` just always use the most recent nightly build?
Might be easier? (if it is allowed to fail)

parsonsmatt:
You must specify a specific resolver in the stack.yaml. You can use `--resolver nightly` to pick the latest. Which I could put in the CI script!

parsonsmatt:
Thanks for the suggestion @Vlix :smile:

PR #212, opened 2020-10-22
Lateral joins

### LINKED ISSUE #121: Postgres JSON joining
If you've got non-trivial JSON in your codebase, it is common to want to `JOIN` against it.

```sql
SELECT foo.id, array_vals.blah
FROM foo
CROSS JOIN jsonb_array_elements(foo.jsonb_col -> 'json_array') AS array_vals
WHERE foo.id = 3
```

This isn't currently possible with `esqueleto` as far as I can tell.

belevy:

After submitting your PR:

------

Fixes #121 

I am not particularly happy with the extra *Lateral constructors on the From GADT. If you can come up with a better way please do. 
Once again the TypeError's have caused so many headaches. Because they need the catchall overlappable instances the only way I could satisfy the compiler was by making new classes with an extra type parameter that is generated using the IsLateral type family. Once again I am very open to suggestions on how to make this implementation less awful but I like the syntax for the end user.

### DIFF
--- a/src/Database/Esqueleto/Experimental.hs
+++ b/src/Database/Esqueleto/Experimental.hs
@@ -471,33 +471,10 @@ data From a where
                         => From a
                         -> (From b, (ToMaybeT a :& b) -> SqlExpr (Value Bool))
                         -> From (ToMaybeT a :& b)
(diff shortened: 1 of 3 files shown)


belevy:
subsumed by #215

PR #213, opened 2020-10-23
Common table expressions

belevy:

After submitting your PR:

========================================

TODO 
   can we get a more unified syntactic style between the recursive cte `unionAll_`/`union_` and `UnionAll`/`Union`
   documentation
   this may be broken af
@parsonsmatt have fun

### DIFF
--- a/src/Database/Esqueleto/Experimental.hs
+++ b/src/Database/Esqueleto/Experimental.hs
@@ -36,6 +36,7 @@ module Database.Esqueleto.Experimental
     , on
     , from
     , (:&)(..)
+    , with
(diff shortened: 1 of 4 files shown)


belevy:
I think i was a bit overzealous with the type signature of with/Recursive. Its not that they return a full query but rather something that cant be selected from. I am working on updating this to defer to a From a, I also intend to obviate the need for marking subqueries in the join tree by making a ToFrom (SqlQuery a). This will unify the interface between a CTE and a SubQuery

belevy:
subsumed by #215

PR #215, opened 2020-10-25
[Experimental] More powerful queries

### LINKED ISSUE #121: Postgres JSON joining

### LINKED ISSUE #124: Common Table Expressions
It'd be *really* nice to have CTE support in Esqueleto.

```haskell
select 
  $ with (from $ \u -> where_ (p u) >> pure u) 
  $ \x -> from 
  $ \y -> -- etc..
```

### LINKED ISSUE #207: No docs for 3.3.4.0 on Hackage
There are no docs for esqueleto 3.3.4.0 on Hackage - https://hackage.haskell.org/package/esqueleto-3.3.4.0

Is this the right place to be reporting this? Sorry if it isn't!

belevy:

After submitting your PR:

This subsumes #213 and #212. It is the merger of the two. Like i mentioned over in #212 I am not particularly happy with the need to maintain so much machinery to support LATERAL but without the newtype wrapper I rather like the syntax here.
```
Experimental.from $ Table @Lord 
`InnerJoin` (\lords -> do
    ...)
`on` (\(l :& r) -> ...)
```

Further I am still unhappy with the fact that `union_` and `unionAll_` feel out of place in the withRecursive syntax compared to the SqlSetOperation syntax. 

This PR:
fixes #207
fixes #124 
fixes #121


belevy:
Only 150 more lines of indirection and the two kinds of unions are unified. I am still unhappy with how verbose the code is and it still needs comments. In addition this change obviates the need to ever use `SelectQuery`. `union_` et al can take `SqlQuery a` directly. Likewise `SubQuery` is also not needed as `SqlQuery a` in the join tree will be treated as a sub query and `a -> SqlQuery b` as a lateral query. Both `with` and `withRecursive` return a `SqlQuery (From a)`. The logic is much the same as a `memo` function where you are memoizing a sub query and so it is effectful, if you call it twice you will add two cte's to the query even if they are identical. The resulting from can be used anywhere in a Join Tree and follows the semantics of CTEs in raw sql, namely that you must select from the CTE to get any data. While regular CTE's are useful from a compositional standpoint in raw sql they lose that value in esqueleto as you could just use a function instead. 

@parsonsmatt I think I am done with anything that I can do along this path, the only way forward from here is to remove stuff. I am really not super excited to maintain this implementation and it feels like there must be a simpler way.

src/Database/Esqueleto/Experimental.hs
```diff
@@ -46,44 +61,95 @@ module Database.Esqueleto.Experimental
...
+             , between, (+.), (-.), (/.), (*.)
+             , random_, round_, ceiling_, floor_
+             , min_, max_, sum_, avg_, castNum, castNumM
+             , coalesce, coalesceDefault
+             , lower_, upper_, trim_, ltrim_, rtrim_, length_, left_, right_
+             , like, ilike, (%), concat_, (++.), castString
+             , subList_select, valList, justList
+             , in_, notIn, exists, notExists
+             , set, (=.), (+=.), (-=.), (*=.), (/=.)
+             , case_, toBaseId
```
jkachmar, line 79: This block has some weird indentation; is that intentional or a copy/paste formatting error from another module?

belevy: Looks like copy pasta

charukiewicz: I copied this directly from `Database.Esqueleto`. I think the original author's reasoning was that these are more or less all operators that go inside of a `where_`. The additional indent can be removed but I don't recommend flattening the entire list, as that would add like 60 lines to the export list.

parsonsmatt: That's from when the class used to be a thing, and it was explicitly exported, so it was like

```
module Database.Esqueleto 
  ( class Esquleto ( subList_select, in_, notIn, exists, notExists, ...)
  ,   -- etc
  )
```

Rather than flattening the export list I just deleted the `class Esqueleto` bit and surrounding parens.
Laziness more than any principled thing.

src/Database/Esqueleto/Internal/PersistentImport.hs
```diff
@@ -138,6 +138,7 @@ module Database.Esqueleto.Internal.PersistentImport
   SqlType(..),
   UniqueDef(..),
   UpdateException(..),
+  -- * Here's a thing
```
parsonsmatt, line 141: ???

test/Common/Test.hs
```diff
@@ -891,7 +891,7 @@ testSelectSubQuery run = do
         let q = do
                 p <- Experimental.from $ Table @Person
                 return ( p ^. PersonName, p ^. PersonAge)
-        ret <- select $ Experimental.from $ SelectQuery q
+        ret <- select $ Experimental.from q
```
parsonsmatt, line 894: yesssssssssssssssss

test/Common/Test.hs
```diff
@@ -991,10 +991,9 @@ testSelectSubQuery run = do
         l3k <- insert l3
         let q = do
                 (lord :& (_, dogCounts)) <- Experimental.from $ Table @Lord
-                        `LeftOuterJoin` (SelectQuery $ do
-                                      lord <- Experimental.from $ Table @Lord
-                                      pure (lord ^. LordId, lord ^. LordDogs)
-                                   )
+                        `LeftOuterJoin` do
+                            lord <- Experimental.from $ Table @Lord
+                            pure (lord ^. LordId, lord ^. LordDogs)
```
parsonsmatt, line 996: image

test/PostgreSQL/Test.hs
```diff
@@ -1151,6 +1153,146 @@ testFilterWhere =
...
+
+        void $ select $ do
+          limitedLordsCte <-
+            Experimental.with $ do
+              lords <- Experimental.from $ Experimental.Table @Lord
+              limit 10
+              pure lords
+          lords <- Experimental.from limitedLordsCte
+          orderBy [asc $ lords ^. LordId]
+          pure lords
```
parsonsmatt, line 1170: incredible

test/PostgreSQL/Test.hs
```diff
@@ -1151,6 +1153,146 @@ testFilterWhere =
...
+    let
+      oneToTen =
+        Experimental.withRecursive
+           (pure $ val (1 :: Int))
+           Experimental.unionAll_
+           (\self -> do
+               v <- Experimental.from self
+               where_ $ v <. val 10
+               pure $ v +. val 1
+           )
```
parsonsmatt, line 1203: omg

src/Database/Esqueleto/Experimental.hs
```diff
@@ -329,6 +397,56 @@ import GHC.TypeLits
...
+--   AS max_sale,
+--   LATERAL
+--   (SELECT customer_name
+--     FROM all_sales
+--     WHERE all_sales.salesperson_id = salesperson.id
+--     AND all_sales.amount =
+--         -- the cached maximum size
+--         max_sale.amount)
+--   AS max_sale_customer;
+-- @
```
parsonsmatt, line 449: image

src/Database/Esqueleto/Experimental.hs
```diff
@@ -344,91 +462,187 @@ import GHC.TypeLits
...
+    Use the 'except_' function instead of the 'Except' data constructor" #-}
+data Except a b = a `Except` b
+
+-- | @EXCEPT@ SQL set operation. Can be used as an infix function between 'SqlQuery' values.
+except_ :: a -> b -> Except a b
+except_ = Except
+
+{-# DEPRECATED Intersect "/Since: 3.4.0.0/ - \
+    Use the 'intersect_' function instead of the 'Intersect' data constructor" #-}
+data Intersect a b = a `Intersect` b
```
parsonsmatt, line 533: good migration patterns?!

src/Database/Esqueleto/Experimental.hs
```diff
@@ -344,91 +462,187 @@ import GHC.TypeLits
...
-                -> From (a :& ToMaybeT b)
-  RightJoinFrom :: ToMaybe a
-                => From a
-                -> (From b, (ToMaybeT a :& b) -> SqlExpr (Value Bool))
-                -> From (ToMaybeT a :& b)
-  FullJoinFrom  :: (ToMaybe a, ToMaybe b )
-                => From a
-                -> (From b, (ToMaybeT a :& ToMaybeT b) -> SqlExpr (Value Bool))
-                -> From (ToMaybeT a :& ToMaybeT b)
+  Table                 :: PersistEntity ent => From (SqlExpr (Entity ent))
```
parsonsmatt, line 576: Formatting suggestion: instead of right-aligning on the longest constructor name, indent the type sig further

eg

```haskell
data From a where
  Table 
    :: PersistEntity ent => From (SqlExpr (Entity ent))
  SubQuery
    :: ( constraints )
    => SqlQuery a
    -> From a''
    -> From a
-- etc
```

This will make for better diffs in the future.

src/Database/Esqueleto/Experimental.hs
```diff
@@ -440,26 +654,56 @@ data From a where
 -- \`on\` (\\(p :& bP) ->
 --         p ^. PersonId ==. bP ^. BlogPostAuthorId)
 -- @
---
-on :: ToFrom a => a -> (b -> SqlExpr (Value Bool)) -> (a, b -> SqlExpr (Value Bool))
+class ValidOnClauseValue a where
```
parsonsmatt, line 657: It seems like this is just a replacement for `ToFrom` - can you make a note on what it's all about? Should an end user ever be defining this?

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -137,12 +139,6 @@ where_ expr = Q $ W.tell mempty { sdWhereClause = Where expr }
 -- and tuple-joins do not need an 'on' clause, but 'InnerJoin' and the various
 -- outer joins do.
 --
--- Note that this function will be replaced by the one in
--- "Database.Esqueleto.Experimental" in version 4.0.0.0 of the library. The
--- @Experimental@ module has a dramatically improved means for introducing
--- tables and entities that provides more power and less potential for runtime
--- errors.
```
parsonsmatt, line 144: Why is this removed?

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1410,12 +1406,6 @@ class ToBaseId ent where
 
 -- | @FROM@ clause: bring entities into scope.
 --
--- Note that this function will be replaced by the one in
--- "Database.Esqueleto.Experimental" in version 4.0.0.0 of the library. The
--- @Experimental@ module has a dramatically improved means for introducing
--- tables and entities that provides more power and less potential for runtime
--- errors.
```
parsonsmatt, line 1417: Why is this removed?

parsonsmatt (approved):
Love it. A few minor questions and doc requests.

src/Database/Esqueleto/Internal/Internal.hs
charukiewicz: @parsonsmatt This was removed when we were trying to figure out why docs weren't building in 3.3.4.0. My bad, should be re-added.

src/Database/Esqueleto/Experimental.hs
```diff
@@ -573,76 +573,99 @@ pattern SelectQuery q = SelectQueryP Never q
...
+    -> From (ToMaybeT a :& b)
+  FullJoinFrom
+    :: (ToMaybe a, ToMaybe b )
+    => From a
+    -> (From b, (ToMaybeT a :& ToMaybeT b) -> SqlExpr (Value Bool))
+    -> From (ToMaybeT a :& ToMaybeT b)
+
+-- | Constraint for `on`. Ensures that only types that require an `on` can be used on
+-- the left hand side. This was previously reusing the ToFrom class which was actually
+-- a bit too lenient as it allowed to much. Expanding this class should not be needed.
```
parsonsmatt, line 662:
```suggestion
-- a bit too lenient as it allowed to much. Expanding this class should not be needed.
--
-- @since 3.4.0.0
```

src/Database/Esqueleto/Experimental.hs
belevy: this will be fixed with #218

PR #220, opened 2020-10-29
Use type families instead of empty classes

googleson78:
An idea I had while looking through the changes. Not sure if this breaks compatibility with a ghc, but it seems more "minimalistic" and I think it shouldn't be a breaking API change.

@belevy


parsonsmatt (approved):
I like this, should be an improvement. @belevy any thoughts?

belevy:
hah, only reason I didn't do this is because I couldn't be bothered to find out where to import Constraint from

parsonsmatt:
Thanks! :smile:

PR #224, opened 2020-10-29
CI Improvements

parsonsmatt:


bitemyapp:
>overhead from starting the jobs is much higher than is saved by parallelizing them

classic CI problem that keeps people on grody solutions like Jenkins

PR #221, opened 2020-10-29
Deprecate ToAliasT and ToAliasReferenceT

arthurxavierx:
Before submitting your PR, check that you've:

- [x] Bumped the version number

After submitting your PR:

- [x] Update the Changelog.md file with a link to your PR
- [ ] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts)

---

This PR deprecates the `ToAliasT` and `ToAliasReferenceT` type families, as both are just the identity on types. This makes it possible for library users to define their own `ToAlias` and `ToAliasReference` instances, which is particularly useful when you're using a custom data type (with a `SqlSelect` instance) in queries.

~~I'm not entirely sure of why we need `ToAliasT` and `ToAliasReferenceT` if in all cases `ToAliasT a ~ a` and `ToAliasReferenceT a ~ a`; maybe they could be removed?~~

Resolves #211

PR #225, opened 2020-10-29
Simplify ToFromT

### LINKED ISSUE #222: Refactor ToFromT
Currently ToFromT does too much. In addition to providing the result type of a ToFrom instance it also is handling type errors for the different join kinds. This prevents it from becoming an associated type family. By extracting out the type error parts we can associate it. 

Already started work on this at https://github.com/foxhound-systems/esqueleto/tree/simplify-type-families

belevy:
Simplify ToFromT moving the overlapping bits out and into their own closed type families with a single purpose. This seems like positive change. 
Make ToMaybeT into an associated type family in the vein of #221 

fix #222

After submitting your PR:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,5 @@ stack.yaml.lock
 .cabal-sandbox/
 cabal.sandbox.config
 .hspec-failures
+*.swp
+*.swo
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,3 @@ stack.yaml.lock
 .cabal-sandbox/
 cabal.sandbox.config
 .hspec-failures
-*.swp
-*.swo
--- a/src/Database/Esqueleto/Experimental.hs
+++ b/src/Database/Esqueleto/Experimental.hs
@@ -753,6 +753,7 @@ class ToFrom a where
     type ToFromT a
     toFrom :: a -> From (ToFromT a)
 
+-- @since 3.4.0.1
 type family FromOnClause a where
     FromOnClause (a, b -> SqlExpr (Value Bool)) = b
     FromOnClause a = TypeError ('Text "Missing ON clause")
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -5,3 +5,5 @@ stack.yaml.lock
 .cabal-sandbox/
 cabal.sandbox.config
 .hspec-failures
+*.swp
+*.swo
```
parsonsmatt, line 9: fwiw this should go in your `~/.gitconfig` : https://git-scm.com/docs/gitignore

belevy:
@parsonsmatt did i get the styling right?

src/Database/Esqueleto/Experimental.hs
```diff
@@ -779,103 +750,149 @@ type family ErrorOnLateral a :: Constraint where
...
+    type ToFromT (LeftOuterJoin a b) = FromOnClause b
     toFrom = undefined
-instance {-# OVERLAPPABLE #-} ToFrom (RightOuterJoin a b) where
-   toFrom = undefined
 instance {-# OVERLAPPABLE #-} ToFrom (FullOuterJoin a b) where
+    type ToFromT (FullOuterJoin a b) = FromOnClause b
+    toFrom = undefined
+instance {-# OVERLAPPABLE #-} ToFrom (RightOuterJoin a b) where
+    type ToFromT (RightOuterJoin a b) = FromOnClause b
+    toFrom = undefined
```
parsonsmatt, line 771: These `undefined`s are kinda weirding me out. If we're not ever calling them, should we delete the method? If we are calling them, shouldn't we implement these?

parsonsmatt:
Formatting looks good

src/Database/Esqueleto/Experimental.hs
belevy: These instances are the defaults that throw the type error. They are undefined to suppress the warning caused by not defining ToFrom.

belevy: If you have a better implementation I am not stoked about how the Type errors work. But it's still better than what was there before in my mind

parsonsmatt: Ooh, I see. It's a bit indirected, but it makes sense.

parsonsmatt (approved):
i dig it

parsonsmatt:
this is g2g pending those checkboxes

PR #226, opened 2020-10-30
Support persistent-2.11

parsonsmatt:

After submitting your PR:

Pending release of `2.11` on Hackage, of course, but want to ensure that we're tracking changes.


src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -205,17 +206,25 @@ upsertBy
...
+            throw (UnexpectedCaseErr OperationNotSupported)
+        Just upsertSql -> do
+            let (updateText, updateVals) =
+                    updatesText sqlB
+                queryText =
+                    upsertSql entDef uniqueFields updateText
+                queryVals =
+                    addVals updateVals
+            xs <- rawSql queryText queryVals
+            pure (head xs)
```
parsonsmatt, line 222: `connUpsertSql` was changed to take the field names directly. This is the only semantic change in the PR.

src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -205,17 +206,33 @@ upsertBy
...
+                upsertSql entDef uniqueFields updateText
+            queryVals =
+                addVals updateVals
+        xs <- rawSql queryText queryVals
+        pure (head xs)
+#else
+    uDef = toUniqueDef uniqueKey
     handler conn f = fmap head $ uncurry rawSql $
         (***) (f entDef (uDef :| [])) addVals $ updatesText conn
+#endif
```
parsonsmatt, line 235: not a huge fan of CPP but it'd be nice to keep this available to 2.10 for a bit

PR #232, opened 2020-12-04
Also export ValidOnClauseValue

Vlix:
This is a constraint on `on`, but not exported, so you have to go into the source to see what it does. Exporting makes this easier.

(I don't think any `ChangeLog` or version edits are necessary for this)


parsonsmatt:
This exports a new term, which would be a minor change, and thus requires an entry in the changelog and a new version.

parsonsmatt:
fixed the CI stuff in #233 , merge that in and try again?

Vlix:
@parsonsmatt should be fine now? :+1:

parsonsmatt:
thanks so much!

PR #236, opened 2021-01-12
Correct unsafeSqlBinOp to correctly set parens

### LINKED ISSUE #235: WHERE expressions sometimes need parens and don't get them
For example:
```
q :: SqlQuery (Value Int)
q = do
  where_ (subSelectUnsafe (pure (val True)))
  pure (val 0)
```
renders as
```
SELECT ?
WHERE SELECT ?
```

which is a syntax error, at least under Postgres. I commented on #79 already thinking it was related, but after a second look I'm not sure.

If we generated this instead, I think we'd be okay.
```
SELECT ?
WHERE (SELECT ?)
```

I think this might be due to the parens field of `ERaw` being ignored in `makeWhere`:

```haskell
makeWhere :: IdentInfo -> WhereClause -> (TLB.Builder, [PersistValue])
makeWhere _    NoWhere                       = mempty
makeWhere info (Where v) = first ("\nWHERE " <>) $ x info
  where
    x =
        case v of
            ERaw _ f             -> f
            EAliasedValue i _    -> aliasedValueIdentToRawSql i
            EValueReference i i' -> valueReferenceToRawSql i i'
            ECompositeKey _      -> throw (CompositeKeyError -> EsqueletoError
CompositeKeyErr MakeWhereError)
```

belevy:
Fixes
#235 

After submitting your PR:

### DIFF
--- a/src/Database/Esqueleto/Internal/Internal.hs
+++ b/src/Database/Esqueleto/Internal/Internal.hs
@@ -2195,7 +2195,7 @@ unsafeSqlCase when v = ERaw Never buildCase
 -- In the example above, we constraint the arguments to be of the
 -- same type and constraint the result to be a boolean value.
 unsafeSqlBinOp :: TLB.Builder -> SqlExpr (Value a) -> SqlExpr (Value b) -> SqlExpr (Value c)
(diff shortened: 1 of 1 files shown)


belevy:
Need to update the Postgres module to add correct parens

PR #238, opened 2021-02-05
Fix non-exhaustive patterns in 'unsafeSqlAggregateFunction'

arthurxavierx:
Got the following error when using `unsafeSqlAggregateFunction`:
```
An exception has occurred: src/Database/Esqueleto/PostgreSQL.hs:94:30-50: Non-exhaustive patterns in lambda
```

This PR fixes the problem.

---

After submitting your PR:

### DIFF
--- a/src/Database/Esqueleto/Internal/Internal.hs
+++ b/src/Database/Esqueleto/Internal/Internal.hs
@@ -2160,7 +2160,7 @@ unsafeSqlCase when v = ERaw Never buildCase
   where
     buildCase :: IdentInfo -> (TLB.Builder, [PersistValue])
     buildCase info =
-        let (elseText, elseVals) = valueToSql v info
+        let (elseText, elseVals) = valueToRawSqlParens v info
             (whenText, whenVals) = mapWhen when info
         in ( "CASE" <> whenText <> " ELSE " <> elseText <> " END", whenVals <> elseVals)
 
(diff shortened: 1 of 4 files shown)


src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -2172,15 +2172,20 @@ unsafeSqlCase when v = ERaw Never buildCase
...
-    valueToSql (EAliasedValue i _) = aliasedValueIdentToRawSql i
-    valueToSql (EValueReference i i') = valueReferenceToRawSql i i'
+-- | (Internal) Convert a value to a raw SQL builder, preserving parens around
+-- 'ERaw' SQL expressions. This is useful for turning values into function or
+-- operator arguments.
+--
+-- Since: 3.4.0.2
+valueToRawSqlParens :: SqlExpr (Value a) -> IdentInfo -> (TLB.Builder, [PersistValue])
+valueToRawSqlParens (ERaw p f) = (first (parensM p)) . f
+valueToRawSqlParens (ECompositeKey _) = throw (CompositeKeyErr SqlCaseError)
```
arthurxavierx, line 2186: Just noticed that the error type thrown here is different. Any suggestions?

src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -91,7 +91,7 @@ unsafeSqlAggregateFunction name mode args orderByClauses = ERaw Never $ \info ->
                 [] -> ""
                 (_:_) -> " "
         (argsTLB, argsVals) =
-            uncommas' $ map (\(ERaw _ f) -> f info) $ toArgList args
```
parsonsmatt, line 94: Ugh, we even have the right warning on. I need to do a warnings-clean PR on the project soon.

src/Database/Esqueleto/Internal/Internal.hs
arthurxavierx: It seems risky to have this same function repeated and specialized more than once in the codebase. I'll go with making the error an argument here.

parsonsmatt: I really don't like how error handling is done in esqueleto at the moment. Since this is a moving of code, it's probably fine to leave as-is. Ideally we'd just fix the representation so this can't happen, but it's not clear to me the best way to do that :thinking:

belevy:
This actually shouldn't be a problem in my work to unify the representation of SqlExpr

arthurxavierx:
I think this is good to merge, but CI isn't helping me :p

belevy:
@parsonsmatt This is fine. We should push 3.4.1.0 to hackage.

belevy:
looks like this changed the order of JSON objects in the postgresql test suite. @arthurxavierx @parsonsmatt i dont particularly care to fix this as i think 3.5.0 essentially reverts this change.

arthurxavierx:
Thank you @belevy! :)

PR #243, opened 2021-03-26
get persistent 2.12 going

parsonsmatt:

After submitting your PR:


src/Database/Esqueleto/Internal/PersistentImport.hs
```diff
@@ -127,7 +125,6 @@ module Database.Esqueleto.Internal.PersistentImport
     FieldType(..),
     ForeignDef(..),
     ForeignFieldDef,
-    HaskellName(..),
```
parsonsmatt, line 130: Well, this is kind of a breaking change. But since this is only a re-export module... and it's only going to work with persistent-2.12... I'm going to make it a minor bump.

PR #244, opened 2021-03-26
Allow attoparsec 0.14

felixonmars:
Tested to build fine here.


parsonsmatt:
will make hackage revisions, thanks!

PR #252, opened 2021-05-13
Update status badge with GH actions and delete some whitespaces from README

ibarrae:
I don't think any of the checkboxes does apply since it's only the README :smile: but if not, let me know to update them properly


parsonsmatt (approved):
awesome, thanks so much! 😄

PR #265, opened 2021-06-02
Add `selectOne`

### LINKED ISSUE #257: Add `selectFirst` Esqueleto variant
`esqueleto` re-exports the function [selectFirst](https://hackage.haskell.org/package/persistent-2.13.0.2/docs/Database-Persist-Class-PersistQuery.html#v:selectFirst) from `persistent` at the moment, I was wondering if a new function with similar functionality could be introduced. I've seen scenarios in which we only want 1 result or nothing from different queries. Probably doing something similar to the following:

``` haskell
fmap listToMaybe <$>
  E.select $
    E.from $ \table ->
      E.where_ $ ...
      E.limit 1
     return table
```

Something I have in mind (that I'm not even sure it compiles) would be something like the following:
``` haskell
selectFirst :: (SqlSelect a r, MonadIO m) => SqlQuery a -> SqlReadT m (Maybe r)
selectFirst query = do
    limit 1
    select query
````

I wonder if that's possible in `esqueleto`, probably would require some `lifting` though :disappointed_relieved:. I hope all of these makes sense and thanks for the great library!

### PR DESCRIPTION
**Pull request by ibarrae**
Closes #257

After submitting your PR:

### DIFF
--- a/src/Database/Esqueleto.hs
+++ b/src/Database/Esqueleto.hs
@@ -97,7 +97,7 @@ module Database.Esqueleto {-# WARNING "This module will switch over to the Exper
   , SqlExpr
   , SqlEntity
   , select
-  , selectSingle
+  , selectFirst
   , selectSource
   , delete
   , deleteCount
@@ -130,7 +130,7 @@ module Database.Esqueleto {-# WARNING "This module will switch over to the Exper
   ) where
 
 import Database.Esqueleto.Legacy
-import Database.Esqueleto.Internal.PersistentImport
+import Database.Esqueleto.Internal.PersistentImport hiding (selectFirst)
 
 
 -- $setup
--- a/src/Database/Esqueleto.hs
+++ b/src/Database/Esqueleto.hs
@@ -97,6 +97,7 @@ module Database.Esqueleto {-# WARNING "This module will switch over to the Exper
   , SqlExpr
   , SqlEntity
   , select
+  , selectSingle
   , selectSource
   , delete
   , deleteCount
--- a/src/Database/Esqueleto.hs
+++ b/src/Database/Esqueleto.hs
@@ -97,7 +97,7 @@ module Database.Esqueleto {-# WARNING "This module will switch over to the Exper
   , SqlExpr
   , SqlEntity
   , select
-  , selectFirst
+  , selectOne
   , selectSource
   , delete
   , deleteCount
@@ -130,7 +130,7 @@ module Database.Esqueleto {-# WARNING "This module will switch over to the Exper
   ) where
 
 import Database.Esqueleto.Legacy
-import Database.Esqueleto.Internal.PersistentImport hiding (selectFirst)
+import Database.Esqueleto.Internal.PersistentImport
 
 
 -- $setup
--- a/src/Database/Esqueleto/Internal/Internal.hs
+++ b/src/Database/Esqueleto/Internal/Internal.hs
@@ -2539,6 +2539,31 @@ select query = do
     conn <- R.ask
     liftIO $ with res $ flip R.runReaderT conn . runSource
 
+-- | Execute an @esqueleto@ @SELECT@ query inside @persistent@'s
+-- 'SqlPersistT' monad and return the first entry wrapped in a @Maybe@.
+--
+-- === __Example usage__
+--
+-- @
+-- firstPerson :: MonadIO m => SqlPersistT m (Maybe Person)
+-- firstPerson =
+    'selectFirst' $
+        'from' $ \person ->
+            return person
+-- @
+--
+-- The above query is equivalent to a 'select' combined with 'limit' but you
+-- would still have to transform the results from a list:
+--
+-- @
+-- firstPerson :: MonadIO m => SqlPersistT m [Person]
+-- firstPerson =
+    'select' $
+        'from' $ \person -> do
+            'limit' 1
+            return person
+-- @
+
 selectFirst :: (SqlSelect a r, MonadIO m) => SqlQuery a -> SqlReadT m (Maybe r)
 selectFirst query = fmap Maybe.listToMaybe $ select $ limit 1 >> query
 
--- a/src/Database/Esqueleto/Internal/Internal.hs
+++ b/src/Database/Esqueleto/Internal/Internal.hs
@@ -2541,31 +2541,32 @@ select query = do
 
 -- | Execute an @esqueleto@ @SELECT@ query inside @persistent@'s
 -- 'SqlPersistT' monad and return the first entry wrapped in a @Maybe@.
+-- @since 3.5.1.0
 --
 -- === __Example usage__
 --
 -- @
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by belevy** [comment]
I'd prefer that the function be called selectFirst, unfortunately that would technically be a major API break since we're re-exporting persistent. Perhaps we could put this in the release that removes the old syntax?

**Comment by ibarrae** [comment]
> I'd prefer that the function be called selectFirst, unfortunately that would technically be a major API break since we're re-exporting persistent. Perhaps we could put this in the release that removes the old syntax?

Thanks for the quick reply @belevy :) I was wondering about that, to avoid conflicts with existent persistent function, thus I thought as a workaround (for now) to get it called in a different way.

And correct me if I'm wrong, but you are mentioning to put this as part of `4.0.0.0` right?

**Comment by belevy** [comment]
@parsonsmatt Correct me if I'm wrong but this would require a bump in the major version right i.e 4.0.0

**Comment by parsonsmatt** [comment]
`esqueleto` uses PVP versioning so we could put this out with `selectFirst` as an `esqueleto` function with `3.6.0.0`

File: src/Database/Esqueleto.hs [hunk 1]
```diff
@@ -129,7 +130,7 @@ module Database.Esqueleto {-# WARNING "This module will switch over to the Exper
   ) where
 
 import Database.Esqueleto.Legacy
-import Database.Esqueleto.Internal.PersistentImport
+import Database.Esqueleto.Internal.PersistentImport hiding (selectFirst)
```
parsonsmatt, line 133: Instead of hiding the import, we should just not re-export it from `PersistentImpot`

src/Database/Esqueleto/Legacy.hs
```diff
@@ -130,7 +131,7 @@ module Database.Esqueleto.Legacy
   ) where
 
 import Database.Esqueleto.Internal.Internal
-import Database.Esqueleto.Internal.PersistentImport
+import Database.Esqueleto.Internal.PersistentImport hiding (selectFirst)
```
parsonsmatt, line 134: Same issue here (and elsewhere) - let's keep this an open import and hide the export at the source.
```suggestion
import Database.Esqueleto.Internal.PersistentImport
```

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -2539,6 +2539,34 @@ select query = do
...
+-- firstPerson :: MonadIO m => SqlPersistT m [Person]
+-- firstPerson =
+--  'select' $
+--      'from' $ \person -> do
+--          'limit' 1
+--          return person
+-- @
+
+selectFirst :: (SqlSelect a r, MonadIO m) => SqlQuery a -> SqlReadT m (Maybe r)
+selectFirst query = fmap Maybe.listToMaybe $ select $ limit 1 >> query
```
parsonsmatt, line 2568: I also like the idea of naming this function `selectFirst`, but the API break with `persistent` does unfortunately require a major version bump (to 3.6.0.0).

It's such a useful function, I'd like to expose it now. We can do this with a minor version bump (3.5.1.0), as long as it does not conflict. `selectOne` is another fine name for it, and I'd honestly be fine with having that name.

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -2539,6 +2539,34 @@ select query = do
...
+-- The above query is equivalent to a 'select' combined with 'limit' but you
+-- would still have to transform the results from a list:
+--
+-- @
+-- firstPerson :: MonadIO m => SqlPersistT m [Person]
+-- firstPerson =
+--  'select' $
+--      'from' $ \person -> do
+--          'limit' 1
+--          return person
```
parsonsmatt, line 2564: Let's use experimental syntax for the new examples.

```suggestion
--  'select' $ do
--      person <- 'from' $ table @Person
--      'limit' 1
--      return person
```

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -2539,6 +2539,34 @@ select query = do
...
+-- 'SqlPersistT' monad and return the first entry wrapped in a @Maybe@.
+--
+-- === __Example usage__
+--
+-- @
+-- firstPerson :: MonadIO m => SqlPersistT m (Maybe Person)
+-- firstPerson =
+--  'selectFirst' $
+--      'from' $ \person ->
+--          return person
```
parsonsmatt, line 2552: Same as below - let's use the `Experimental` syntax instead, since all code will be switching to it.

parsonsmatt (changes requested):
I'd rather have this out now with `selectOne` as the name than do a major version bump to give it the name `selectFirst`.

This is great to have, thanks for the PR! 😄

ibarrae:
Thanks for all the suggestions @belevy @parsonsmatt, I updated the branch accordingly and also updated the version number. Any more feedback is more than welcome :tada:

test/Common/Test.hs
```diff
@@ -297,6 +297,27 @@ testSubSelect = do
                 Right xs ->
                     xs `shouldBe` []
 
+testSelectOne :: SpecDb
+testSelectOne =
+    describe "selectOne" $ do
+        let personQuery =
+                selectOne $
+                    from $ \person -> do
```
belevy, line 305: could you change this test query to use the Experimental syntax please. Were in the process of deprecating the old syntax so it will eventually need to change anyway.

ibarrae: For sure @belevy, I still need to get used to the upcoming syntax :smile:. You can take a look at those changes in this [commit](https://github.com/bitemyapp/esqueleto/pull/265/commits/7622e6881491960776f6f682414c3e6659adeb44)

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -2539,6 +2539,35 @@ select query = do
...
+-- The above query is equivalent to a 'select' combined with 'limit' but you
+-- would still have to transform the results from a list:
+--
+-- @
+-- firstPerson :: MonadIO m => SqlPersistT m [Entity Person]
+-- firstPerson =
+--  'select' $ do
+--      person <- 'from' $ 'table' @Person
+--      'limit' 1
+--      return person
```
parsonsmatt, line 2565: fantastic 😄

test/Common/Test.hs
```diff
@@ -297,6 +297,27 @@ testSubSelect = do
...
+            person <- insert' p1
+            _ <- insert' p2
+            res <- personQuery
+            asserting $
+                res `shouldBe` Just (Value $ entityKey person)
+
+        itDb "returns Nothing" $ do
+            res <- personQuery
+            asserting $
+                res `shouldBe` (Nothing :: Maybe (Value PersonId))
```
parsonsmatt, line 319: awesome

parsonsmatt (approved):
looks great, thank you 😄

belevy (approved):

parsonsmatt:
released! thanks 😄

Issue #256, opened 2021-05-26
Add pgcrypto functions 


3kyro:
Hi, I have a local implementation of the following functions from the [pgcrypto](https://www.postgresql.org/docs/8.3/pgcrypto.html) module:

```` haskell
toCrypt :: SqlString s => Algorithm -> s -> SqlExpr (Value s)
fromCrypt :: SqlString s => SqlExpr (Value s) -> s -> SqlExpr (Value Bool)
````
Usage with a contrived example would be:

```` haskell
-- insert a User
insertSelect $
    User
      <# val "new_user"
      <&> toCrypt BF "1234Password"
      -- etc

-- validate a User
select $ do
    user <- from $ Table @User
    where_ $ user ^. UserName ==. val "new_user"
      &&. fromCrypt (user ^. UserPassword) "1234Password"
    pure user
````
I was wondering if they can be included in *esqueleto* - in the PostgreSQL module - or it's better of to create an external library. 

If inclusion to *esqueleto* is OK I'd happily submit a PR

3kyro:
Actually going through #228, I'll wait for it to drop since I am pattern matching against the ``SqlExpr`` constructors.

belevy:
@3kyro sorry about the change causing rework on your side. Id be happy to see a PR for the crypto functions.

3kyro:
@belevy No problem at all. Thanks for the confirmation, I'll submit a PR for this shortly

3kyro:
I'm closing this issue as there was no agreement on the proposed API.

PR #266, opened 2021-06-07
Add pgcrypto crypt() functions

3kyro:
As described in #256

- [x ] Bumped the version number.
- [ x] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html).
- [ x] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock.
- [ x] Ran `stylish-haskell` and otherwise adhered to the [style guide](https://github.com/bitemyapp/esqueleto/blob/master/style-guide.yaml).

After submitting your PR:

- [ x] Update the Changelog.md file with a link to your PR.
- [ x] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts).

### DIFF
--- a/src/Database/Esqueleto/PostgreSQL/Pgcrypto.hs
+++ b/src/Database/Esqueleto/PostgreSQL/Pgcrypto.hs
@@ -1,0 +1,99 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+-- | This module contains functions specific to the [pgcrypto](https://www.postgresql.org/docs/current/pgcrypto.html) module
+-- 
+-- @since: 3.5.1
+
+module Database.Esqueleto.PostgreSQL.Pgcrypto 
+    (HashAlgorithm(..),
+     toCrypt,
+     fromCrypt,
+    ) where
+
+import Database.Esqueleto.Internal.Internal 
+import Database.Esqueleto.Experimental (toPersistValue)
+
+{- pgcrypto hashing algorithms
+see: https://www.postgresql.org/docs/current/pgcrypto.html
+
+/Requires/ the pgcrypto module.
+
+@since 3.5.1
+-}
+data HashAlgorithm = BF | MD5 | XDES | DES
+  deriving (Eq, Show)
+
+{- | (@crypt()@) Calculate a crypt-like hash from the provided password
+
+/Requires/ the pgcrypto module.
+
+/WARNING/: Using `toCrypt` may leak sensitive data via logging. Filtering logs in production environments
+when using `toCrypt`, such as using `filterLogger` on `monad-logger` based stacks is highly advised.
+
+example:
+@
+share
+    [mkPersist sqlSettings]
+    [persistLowerCase|
+    UserAccount json
+        name T.Text
+        UniqueName name
+        passwordHash T.Text
+        deriving Show Read Eq
+
+insertSelect $ do
+    pure $
+        UserAccount
+            <# val "username"
+            <&> toCrypt BF "1234password"
+@
+
+-- @since 3.5.1
+-}
+toCrypt :: SqlString s => HashAlgorithm -> s -> SqlExpr (Value s)
+toCrypt algorithm pass =
+    let alg = case algorithm of
+            BF -> "bf"
+            MD5 -> "md5"
+            XDES -> "xdes"
+            DES -> "des"
+    in ERaw noMeta $ \_ _ -> ("crypt (?, gen_salt('" <> alg <> "'))", [toPersistValue pass])
+
+{- | (@crypt()@) Retrieve a hashed password
+
+/Requires/ the pgcrypto module.
+
+example:
+@
+share
+    [mkPersist sqlSettings]
+    [persistLowerCase|
+    UserAccount json
+        name T.Text
+        UniqueName name
+        passwordHash T.Text
+        deriving Show Read Eq
+
+
+login name pwd = select $ do
+    user <- from $ Table UserAccount
+    where_ $ user ^. UserAccountName ==. val name
+        &&. fromCrypt (user ^. UserAccountPasswordHash) pwd
+    pure user
+@
+
+-- @since 3.5.1
+-}
+fromCrypt :: SqlString s => SqlExpr (Value s) -> s -> SqlExpr (Value Bool)
+fromCrypt expr pass =
+    expr
+        ==. ERaw
+            noMeta
+            ( \_ info ->
+                let name = columnName expr info
+                in ("crypt (?, " <> name <> ")", [toPersistValue pass])
+            )
+  where
+    columnName (ERaw _ f) info =
+            fst $ f Never info
--- a/src/Database/Esqueleto/PostgreSQL/Pgcrypto.hs
+++ b/src/Database/Esqueleto/PostgreSQL/Pgcrypto.hs
@@ -2,26 +2,35 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 
 -- | This module contains functions specific to the [pgcrypto](https://www.postgresql.org/docs/current/pgcrypto.html) module
--- 
+--
 -- @since: 3.5.1
 
-module Database.Esqueleto.PostgreSQL.Pgcrypto 
+module Database.Esqueleto.PostgreSQL.Pgcrypto
     (HashAlgorithm(..),
      toCrypt,
      fromCrypt,
     ) where
 
-import Database.Esqueleto.Internal.Internal 
+import qualified Data.Text.Internal.Builder as TLB
 import Database.Esqueleto.Experimental (toPersistValue)
+import Database.Esqueleto.Internal.Internal
 
 {- pgcrypto hashing algorithms
 see: https://www.postgresql.org/docs/current/pgcrypto.html
+
+`bf` and `xdes` algorithms have an optional iterations count parameter. All limitations and considerations
+mentioned in the `pgcrypto` module documentation regarding iteration count apply. It is possible to supply
+an invalid iteration count, which will lead to an sql error.
 
 /Requires/ the pgcrypto module.
 
 @since 3.5.1
 -}
-data HashAlgorithm = BF | MD5 | XDES | DES
+data HashAlgorithm
+    = BF (Maybe Word)
+    | MD5
+    | XDES (Maybe Word)
+    | DES
   deriving (Eq, Show)
 
 {- | (@crypt()@) Calculate a crypt-like hash from the provided password
@@ -46,7 +55,7 @@ insertSelect $ do
     pure $
         UserAccount
             <# val "username"
-            <&> toCrypt BF "1234password"
+            <&> toCrypt (BF Nothing) "1234password"
 @
 
 -- @since 3.5.1
@@ -54,11 +63,19 @@ insertSelect $ do
 toCrypt :: SqlString s => HashAlgorithm -> s -> SqlExpr (Value s)
 toCrypt algorithm pass =
     let alg = case algorithm of
-            BF -> "bf"
-            MD5 -> "md5"
-            XDES -> "xdes"
-            DES -> "des"
-    in ERaw noMeta $ \_ _ -> ("crypt (?, gen_salt('" <> alg <> "'))", [toPersistValue pass])
+            BF mIterCount ->
+                "'bf'" <> case mIterCount of
+                             Nothing -> mempty
+                             Just iterCount ->
+                                "," <> TLB.fromString (show iterCount)
+            MD5 -> "'md5'"
+            XDES mIterCount ->
+                "'xdes'" <> case mIterCount of
+                             Nothing -> mempty
+                             Just iterCount ->
+                                "," <> TLB.fromString (show iterCount)
+            DES -> "'des'"
+    in ERaw noMeta $ \_ _ -> ("crypt (?, gen_salt(" <> alg <> "))", [toPersistValue pass])
 
 {- | (@crypt()@) Retrieve a hashed password
 
--- a/test/PostgreSQL/Test.hs
+++ b/test/PostgreSQL/Test.hs
@@ -1273,7 +1273,7 @@ testCrypt = do
             _ <- insertSelect $ do
                 pure $ User
                     <# val "name"
-                    <&> PGC.toCrypt PGC.BF "1234password"
+                    <&> PGC.toCrypt (PGC.BF $ Just 6) "1234password"
             authenticated <-
                 select $ do
                         user' <- Experimental.from $ table @User
--- a/test/PostgreSQL/Test.hs
+++ b/test/PostgreSQL/Test.hs
@@ -40,7 +40,7 @@ import Database.Esqueleto.PostgreSQL (random_)
 import qualified Database.Esqueleto.PostgreSQL as EP
 import Database.Esqueleto.PostgreSQL.JSON hiding ((-.), (?.), (||.))
 import qualified Database.Esqueleto.PostgreSQL.JSON as JSON
-import Database.Persist.Postgresql (withPostgresqlConn, createPostgresqlPool)
+import Database.Persist.Postgresql (createPostgresqlPool, withPostgresqlConn)
 import Database.PostgreSQL.Simple (ExecStatus(..), SqlError(..))
 import System.Environment
 import Test.Hspec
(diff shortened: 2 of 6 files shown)


3kyro:
Some comments:

- I have not bumped the version number, will it be 3.5.1.0?
- For the tests,  the esqutest database should have the pgcrypto module installed. Could this simply be put as an additional step in the readme or do I create a migration step?

belevy:
Could you put the functions in their own module. Is pgcrypto installed by default in the github actions container or does that need to be added? Perhaps we should be able to run without the crypto functions?

3kyro:
I think the easiest would be to have a `rawSql` call inside the test

````haskell
rawSql "CREATE EXTENSION IF NO EXISTS pgcrypto" []
````

I'll put the function in their own module.

**Comment by 3kyro** [comment]
> 
> 
> Could you put the functions in their own module. Is pgcrypto installed by default in the github actions container or does that need to be added? Perhaps we should be able to run without the crypto functions?

Another way would be to query `pg_extension` during the test and abort if `pgcrypto` is not found. Would you prefer this rather than enabling it each time?

File: src/Database/Esqueleto/PostgreSQL/Pgcrypto.hs [hunk 1]
```diff
@@ -0,0 +1,99 @@
...
+        &&. fromCrypt (user ^. UserAccountPasswordHash) pwd
+    pure user
+@
+
+-- @since 3.5.1
+-}
+fromCrypt :: SqlString s => SqlExpr (Value s) -> s -> SqlExpr (Value Bool)
+fromCrypt expr pass =
+    expr
+        ==. ERaw
```
belevy, line 91: why is this including ==.

Could we instead just expose crypt and gen_salt as functions?

3kyro: Hi, this comes mostly from the fact that I can't replicate pg's `crypt()` overloaded behavior in haskell in order to only have  `crypt` and `genSalt` functions. A function 

````haskell
crypt :: SqlString s => s -> SqlExpr (Value s) -> SqlExpr (Value s)
````
would need a context switching parameter to identify when it is used for storing passwords and when for retrieving passwords.
 
With this in mind, `crypt()` always needs to use `gen_salt()` when used to store passwords, and so I think it makes sense to embed it in `toCrypt` instead of always including it explicitly. Moreover it protects the user from falsely using always the same salt by passing something like `val "salt"`.  

`fromCrypt` could skip the `==.` but then every time it was used, the user would have to duplicate the same persist field:

```haskell
login name pwd = select $ do
   user <- from $ Table UserAccount
   where_ $ user ^. UserAccountName ==. val name
   -- ^ as before
         &&. user ^. UserAccountPasswordHash ==.  fromCrypt pwd  (user ^. UserAccountPasswordHash)
````
Having the `==.` inside `fromCrypt` makes its use a bit more concise. 

Having said that, I understand that this might not be generic enough for esqueleto, so I would be happy to try to expose genSalt as a function and change `fromCrypt` if you think it would be better for the library.

btw, I've seen that I missed some parameters in some hashing algorithms. I'll add these as well.

Thanks for the help

**Comment by 3kyro** [comment]
> 
> 
> can you update the changelog and version for this.
> 
> @parsonsmatt im still not sure how the numbering scheme works, would this be 3.5.1.0 or 3.5.0.1?

I guess if we follow Haskell PVP this is a non breaking change that adds functionality so 3.5.1.0 is appropriate

> Non-breaking change. Otherwise, if only new bindings, types, classes, non-orphan instances or modules (but see below) were 
> added to the interface, then A.B MAY remain the same but the new C MUST be greater than the old C. Note that modifying     > imports or depending on a newer version of another package may cause extra non-orphan instances to be exported and thus  > force a minor version change."

**Comment by 3kyro** [comment]
Hi @belevy , @parsonsmatt . Is there something I can do to have this merged? Thanks!

File: src/Database/Esqueleto/PostgreSQL/Pgcrypto.hs [hunk 2]
```diff
@@ -0,0 +1,116 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+-- | This module contains functions specific to the [pgcrypto](https://www.postgresql.org/docs/current/pgcrypto.html) module
+--
+-- @since: 3.5.1
```
parsonsmatt, line 6: The `:` is unnecessary in since annotations

```suggestion
-- @since 3.5.1
```

src/Database/Esqueleto/PostgreSQL/Pgcrypto.hs
```diff
@@ -0,0 +1,116 @@
...
+
+module Database.Esqueleto.PostgreSQL.Pgcrypto
+    (HashAlgorithm(..),
+     toCrypt,
+     fromCrypt,
+    ) where
```
parsonsmatt, line 12: The repo uses comma-first style

```suggestion
    ( HashAlgorithm(..)
    , toCrypt
    , fromCrypt
    ) where
```

src/Database/Esqueleto/PostgreSQL/Pgcrypto.hs
```diff
@@ -0,0 +1,116 @@
...
+
+import qualified Data.Text.Internal.Builder as TLB
+import Database.Esqueleto.Experimental (toPersistValue)
+import Database.Esqueleto.Internal.Internal
+
+{- pgcrypto hashing algorithms
```
parsonsmatt, line 18: To make this a doc comment, you'd write 

```suggestion
{- | pgcrypto hashing algorithms
```

src/Database/Esqueleto/PostgreSQL/Pgcrypto.hs
parsonsmatt: As a naming convention, I'd probably want `fromCrypt` to perform the decryption, and another function `checkEncrypted` that does the helper stuff. So your code example would be:

```haskell
    && user ^. UserAccountPasswordHash ==. fromCrypt pwd (user ^. UserAccountPasswordHash)
   -- or
    && encryptedEquals (user ^. UserAccountPasswordHash) pwd
```

src/Database/Esqueleto/PostgreSQL/Pgcrypto.hs
```diff
@@ -0,0 +1,116 @@
...
+login name pwd = select $ do
+    user <- from $ Table UserAccount
+    where_ $ user ^. UserAccountName ==. val name
+        &&. fromCrypt (user ^. UserAccountPasswordHash) pwd
+    pure user
+@
+
+-- @since 3.5.1
+-}
+fromCrypt :: SqlString s => SqlExpr (Value s) -> s -> SqlExpr (Value Bool)
```
parsonsmatt, line 105: The idiom in the library to write functions accepting `SqlExpr (Value s)` instead of `s` directly. You can easily go from `s -> SqlExpr (Value s)` via `val` but you *can't* go from a `SqlExpr (Value s) -> s` without actually running the database query. This is more flexible and powerful.

src/Database/Esqueleto/PostgreSQL/Pgcrypto.hs
```diff
@@ -0,0 +1,116 @@
...
+
+
+login name pwd = select $ do
+    user <- from $ Table UserAccount
+    where_ $ user ^. UserAccountName ==. val name
+        &&. fromCrypt (user ^. UserAccountPasswordHash) pwd
+    pure user
+@
+
+-- @since 3.5.1
```
parsonsmatt, line 103: Comment mark is unnecessary since we're in a block comment.

```suggestion
@since 3.5.1
```

src/Database/Esqueleto/PostgreSQL/Pgcrypto.hs
```diff
@@ -0,0 +1,116 @@
...
+/Requires/ the pgcrypto module.
+
+@since 3.5.1
+-}
+data HashAlgorithm
+    = BF (Maybe Word)
+    | MD5
+    | XDES (Maybe Word)
+    | DES
+  deriving (Eq, Show)
```
parsonsmatt, line 34:
```suggestion
    deriving (Eq, Show)
```

src/Database/Esqueleto/PostgreSQL/Pgcrypto.hs
```diff
@@ -0,0 +1,116 @@
...
+insertSelect $ do
+    pure $
+        UserAccount
+            <# val "username"
+            <&> toCrypt (BF Nothing) "1234password"
+@
+
+-- @since 3.5.1
+-}
+toCrypt :: SqlString s => HashAlgorithm -> s -> SqlExpr (Value s)
```
parsonsmatt, line 63: Same comment as below - it's generally best for esqueleto functions to accept `SqlExpr` arguments instead of regular Haskell values, and let folks promote the values themselves.

If we accepted `SqlExpr (Value HashAlgorithm)`, then we'd be able to eg store the hash algorithm in the database along with the hash. [There's at least one stack overflow answer that suggests this as a good idea](https://stackoverflow.com/questions/5975010/should-i-store-the-hash-function-name-in-the-db-when-hashing-passwords), though I don't know enough about security to know for sure.

If the end user wants to hardcode it, then they can just write `val MD5` or whatever.

src/Database/Esqueleto/PostgreSQL/Pgcrypto.hs
```diff
@@ -0,0 +1,116 @@
+{-# LANGUAGE OverloadedStrings #-}
+{-# LANGUAGE ScopedTypeVariables #-}
+
+-- | This module contains functions specific to the [pgcrypto](https://www.postgresql.org/docs/current/pgcrypto.html) module
+--
+-- @since: 3.5.1
+
```
parsonsmatt, line 7:
```suggestion
```

src/Database/Esqueleto/PostgreSQL/Pgcrypto.hs
```diff
@@ -0,0 +1,116 @@
...
+module Database.Esqueleto.PostgreSQL.Pgcrypto
+    (HashAlgorithm(..),
+     toCrypt,
+     fromCrypt,
+    ) where
+
+import qualified Data.Text.Internal.Builder as TLB
+import Database.Esqueleto.Experimental (toPersistValue)
+import Database.Esqueleto.Internal.Internal
+
```
parsonsmatt, line 17: I feel like the module can use a function `enablePgCryptoModule :: SqlPersistT m ()` that does the `rawExecute` stuff to turn it on.

src/Database/Esqueleto/PostgreSQL/Pgcrypto.hs
```diff
@@ -0,0 +1,116 @@
...
+mentioned in the `pgcrypto` module documentation regarding iteration count apply. It is possible to supply
+an invalid iteration count, which will lead to an sql error.
+
+/Requires/ the pgcrypto module.
+
+@since 3.5.1
+-}
+data HashAlgorithm
+    = BF (Maybe Word)
+    | MD5
```
parsonsmatt, line 31: i feel like we should be attaching a `{-# WARNING MD5 "MD5 is considered insecure. Consider switching to another password hashing algorithm." #-}` so that folks don't try to use it. cf [What Is MD5 and Why Is It Considered Insecure?](https://www.section.io/engineering-education/what-is-md5/#:~:text=Unfortunately%2C%20MD5%20has%20been%20cryptographically,not%20be%20used%20for%20anything.&text=It%20is%20always%20recommended%20to,in%20any%20modern%20programming%20framework.)

src/Database/Esqueleto/PostgreSQL/Pgcrypto.hs
```diff
@@ -0,0 +1,116 @@
...
+    expr
+        ==. ERaw
+            noMeta
+            ( \_ info ->
+                let name = columnName expr info
+                in ("crypt (?, " <> name <> ")", [toPersistValue pass])
+            )
+  where
+    columnName (ERaw _ f) info =
+            fst $ f Never info
```
parsonsmatt, line 116: I think this may break in more complex uses of the library. Do we have a test that ensures that eg a `subSelect` can be used here, or a more complex column (eg `( a ^. FooCol ++. b ^. BarCol)`)?

test/PostgreSQL/Test.hs
```diff
@@ -1264,6 +1265,26 @@ testLateralQuery = do
...
+            authenticated <-
+                select $ do
+                        user' <- Experimental.from $ table @User
+                        where_ $ user' ^. UserUsername ==. val "name"
+                            &&. PGC.fromCrypt (user' ^. UserPasswordHash) "1234password"
+                        pure user'
+            let authUser = entityVal $ head authenticated
+            asserting $ do
+                userUsername authUser `shouldBe` "name"
+                userPasswordHash authUser `shouldNotBe` "1234password"
```
parsonsmatt, line 1286: great test, thank you 😄

parsonsmatt (changes requested):
I think we can approach this a bit more from constructing fundamentals. This approach does a lot of `ERaw` manipulation that is already handled by the library.

For example, we can instead define a Haskell variant of the `crypt` function that's as close to the implementation as possible.

```haskell
crypt
    :: (SqlString s)
    => SqlExpr (Value s)
    -> SqlExpr (Value (SaltOr s))
    -> SqlExpr (Value s)
crypt pwd salt = unsafeSqlFunction (pwd, salt)    
```

We need a way to construct the `SaltOr s` values. These would be something like:

```haskell
data SaltOr s = GenSalt HashAlgorithm | RegularValue s
```

We need a `PersistField` and `PersistFieldSql` instance for it.

```haskell
instance PersistField s => PersistFIeld (SaltOr s) where
    toPersistValue x = 
        case x of
            RegularValue s -> 
                toPersistValue s
            GenSalt hashAlg ->
                toPersistValue s 

    fromPersistValue pv = do
        RegularValue <$> fromPersistValue pv

instance ToPersistValue HashAlgorithm where
    toPersistValue hashAlgorithm =
        PersistValueUnescaped $ renderHashAlgorithm hashAlgorithm

    fromPersistValue pv = 
        -- we probably won't ever parse this? but can't hurt to write a parser
```

With these building blocks, we can now rewrite the API you gave as:

```haskell
toCrypt :: SqlString s => HashAlgorithm -> s -> SqlExpr (Value s)
toCrypt hashAlg s = 
    crypt (val s) (val (GenSalt hashAlg))

fromCrypt :: SqlString s => SqlExpr (Value s) -> s -> SqlExpr (Value Bool)
fromCrypt expr s =
    expr ==. crypt expr (val (RegularValue s))
```

Probably also a good idea to write a conversion function like

```haskell
toRegularValue :: SqlString s => SqlExpr (Value s) -> SqlExpr (Value (SaltOr s))
toRegularValue = veryUnsafeCoerceSqlExprValue

genSalt :: SqlExpr (Value HashAlgorithm) -> SqlExpr (Value (SaltOr x))
genSalt = veryUnsafeCoerceSqlExprValue
```

This change makes for a slightly nicer API, too

```haskell
toCrypt hashAlg pass =
    crypt (val pass) (genSalt (val hashAlg))

fromCrypt expr pass =
    expr ==. crypt expr (toRegularValue (val pass))
```

The `RegularValue` name is awful and a better one is probably `CheckValue` or similar to communicate the intent a bit more.

Anyway, I like the idea, and I want to support it, but I think we should reconsider the approach here a bit. Thanks so much for the contribution!

3kyro:
Hi @parsonsmatt , thanks for the great review and sorry for the delay in answering. I agree that your proposed API is much more concise and elegant. I have a slight worry that we allow the user to use the same salt when storing passwords, eg:
````haskell
-- when storing a password
crypt (val pass) (RegularValue (val "a very bad salt"))
````
however this is a inherent limitation of the original `crypt` function and the user can simply be warned about it in the documentation.  Would you be OK with this?

I have rebased to master and will move on with your suggestions shortly

3kyro:
Hi @parsonsmatt, sorry to drag this on, but I have another issue with the proposed API. Pg's `crypt` function annoyingly swaps arguments depending on the type of operation (store/retrieve).

````
// Example of setting a new password:
UPDATE ... SET pswhash = crypt('new password', gen_salt('md5'));
//-----------------------------------------------------^ GenSalt 

//Example of authentication:
SELECT pswhash = crypt('entered password', pswhash) FROM ... ;
//-------------------------^RegularValue
````
Since we're not able to coerce the `SaltOr` value once wrapped in a `SqlExpr` we won't be able to correctly place the arguments in the call to `unsafeSqlFunction`. The way I see it there are three possibilities:
1. Use separate `toCrypt` and `fromCrypt` functions but with all arguments wrapped in a `SqlExpr` This has the benefit of addressing my previous concern about contant salts, but of course isn't so cool as a single `crypt` function
2. Use a single `crypt` function, but with a pure `SalrOr` argument so that we can pattern match on it. Apart from breaking the library's idiom, wrapping the hash algorithm / requested password would be unwieldy.
3. Use a single `crypt` function and try to deduce the context by parsing the `ERaw` constructor. Seems too hacky.

I would go for option no 1 but I would greatly appreciate your opinion.

PR #268, opened 2021-06-14
Add `SqlSelect` instance for `:&`

### LINKED ISSUE #267: `SqlSelect` instances for `a :& b`
Should `a :& b` have a `SqlSelect` instance, based on that of `(a, b)`? This has come up for me when using the experimental API, since I do something like
```haskell
select $ do
  e <- from ... -- Here, the 'from' is polymorphic and user-supplied
  ... -- Using a polymorphic @e :: rec@
  pure e -- Returning @rec@
```
In the above example, where the `from` is a join, `rec` becomes `a :& b`, but since Esqueleto expects you to split the `:&` and return `(a, b)` instead, there's no appropriate instance for running the `select`.

The instance could look something like this:
```haskell
toJoin :: (a, b) -> (a :& b)
toJoin (a, b) = (a :& b)

fromJoin :: (a :& b) -> (a, b)
fromJoin (a :& b) = (a, b)

instance (SqlSelect a ra, SqlSelect b rb) => SqlSelect (a :& b) (ra :& rb) where
    sqlSelectCols esc = sqlSelectCols esc . fromJoin
    sqlSelectColCount = sqlSelectColCount . fromJoinProxy
      where
       fromJoinProxy :: Proxy (a :& b) -> Proxy (a, b)
       fromJoinProxy Proxy = Proxy
    sqlSelectProcessRow = fmap toJoin . sqlSelectProcessRow
```
This is an orphan I'm using to make my code compile, but I don't know if it's the intended way.

ivanbakel:
Fixes #267.

After submitting your PR:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,10 @@
+3.5.2.0
+=======
+- @ivanbakel
+  - [#268](https://github.com/bitemyapp/esqueleto/pull/268)
+        - Added `SqlSelect` instance for `(:&)`, allowing it to be returned from
+          queries just like `(,)` tuples.
+
 3.5.1.0
 =======
 - @ibarrae
--- a/esqueleto.cabal
+++ b/esqueleto.cabal
@@ -1,7 +1,7 @@
 cabal-version: 1.12
 
 name:           esqueleto
-version:        3.5.1.0
+version:        3.5.2.0
 synopsis:       Type-safe EDSL for SQL queries on persistent backends.
 description:    @esqueleto@ is a bare bones, type-safe EDSL for SQL queries that works with unmodified @persistent@ SQL backends.  Its language closely resembles SQL, so you don't have to learn new concepts, just new syntax, and it's fairly easy to predict the generated SQL and optimize it for your backend. Most kinds of errors committed when writing SQL are caught as compile-time errors---although it is possible to write type-checked @esqueleto@ queries that fail at runtime.
                 .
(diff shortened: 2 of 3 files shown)


esqueleto.cabal
```diff
@@ -1,7 +1,7 @@
 cabal-version: 1.12
 
 name:           esqueleto
-version:        3.5.0.0
+version:        3.5.1.0
```
```suggestion
version:        3.5.1.1
```

changelog.md
```diff
@@ -1,3 +1,10 @@
+3.5.1.0
```
```suggestion
3.5.1.1
```

src/Database/Esqueleto/Experimental/From/Join.hs
```diff
@@ -49,6 +49,18 @@ class ValidOnClause a
 instance {-# OVERLAPPABLE #-} ToFrom a a' => ValidOnClause a
 instance ValidOnClause (a -> SqlQuery b)
 
+-- | You may return joined values from a 'select' query - this is
+-- identical to the tuple instance, but is provided for convenience.
+--
+-- @since 3.5.1.0
```
parsonsmatt, line 55:
```suggestion
-- @since 3.5.1.1
```

parsonsmatt (approved):
Approving conditional on merging `master` and updating the version number.

Since we're adding an instance to a class we control, I think it's fine to make this a patch version.

ivanbakel:
>Since we're adding an instance to a class we control, I think it's fine to make this a patch version.

It's your repo, but the PVP does specify that this should be a minor version bump.

parsonsmatt:
I'm sorry, I was mistaken about that. I'll go ahead and merge-on-green and make that change myself. 

Thanks so much for your contribution! 😄

ivanbakel:
Since I had time, I've bumped the version to 3.5.2.

parsonsmatt:
Thank you! I'll get this released today.

Issue #272, opened 2021-07-08
esqueleto-3.5.0.0 generates malformed SQL when notIn is passed an empty list


cdparks:
An `notIn` expression using the legacy API:

```haskell
fetchUsersExcept :: MonadIO m => [Text] -> SqlReadT m [Entity User]
fetchUsersExcept names = select . from $ \users -> do
  where_ $ users ^. UserName `notIn` valList names
  pure users
```

or the experimental API:

```haskell
fetchUsersExcept :: MonadIO m => [Text] -> SqlReadT m [Entity User]
fetchUsersExcept names = select $ do
  users <- from $ table @User
  where_ $ users ^. UserName `notIn` valList names
  pure users
```

will generate a malformed SQL query when passed an empty list:

```haskell
fetchUsersExcept []
```

=>

```plaintext
[Debug#SQL]
SELECT "users"."id", "users"."name"
FROM "users"
WHERE "users"."name" NOT IN (); []
esqueleto-bug: SqlError {sqlState = "42601", sqlExecStatus = FatalError, sqlErrorMsg = "syntax error at or near \")\"", sqlErrorDetail = "", sqlErrorHint = ""}
```

Apparently Postgres (I'm using 10.14) doesn't like `x NOT IN ()`. Sqlite seems to accept this syntax.

It looks to me like `esqueleto` [protects against passing an empty list](https://github.com/bitemyapp/esqueleto/blob/3a12a15d0084877833d1ff2d36d94e1d47587e4b/src/Database/Esqueleto/Internal/Internal.hs#L931) for `in_`, but not for `notIn`. This seems like a behavior change based on [this older commit](https://github.com/bitemyapp/esqueleto/blob/e39c62990e0854e655e55e2ef6f874641efa26cc/src/Database/Esqueleto/Internal/Internal.hs#L949), which used a helper function [`ifNotEmptyList`](https://github.com/bitemyapp/esqueleto/blob/e39c62990e0854e655e55e2ef6f874641efa26cc/src/Database/Esqueleto/Internal/Internal.hs#L2198).

I have [a repo](https://github.com/cdparks/esqueleto-bug) demonstrating this, but it's probably easiest to verify this in your own setup.

belevy:
#273 fixes the issue

PR #273, opened 2021-07-08
Avoid generating an empty list as the left operand to NOT IN

cdparks:
Postgres treats `x in ()` and `x not in ()` as syntax errors. Instead translate:

```plaintext
  x in ()      ⇒  FALSE
  x not in ()  ⇒  TRUE
```

Older versions of esqueleto did both of these, but apparently the latter got lost in the most recent version. See issue #272.

- [x] Bumped the version number. ~~**I'm not sure what the version number should be, looks like esqueleto uses PVP?**~~ I bumped the smallest component to `3.5.2.1` since this is a bugfix
- [ ] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html). **N/A**
- [ ] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock. **N/A**

After submitting your PR:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,9 @@
+3.5.2.1
+=======
+- @cdparks
+  - [#273](https://github.com/bitemyapp/esqueleto/pull/273)
+        - Avoid generating an empty list as the left operand to `NOT IN`.
+
 3.5.2.0
 =======
 - @ivanbakel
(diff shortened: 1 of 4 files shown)


belevy (approved):

cdparks:
@belevy let me know if I bumped the version/updated the changelog correctly

belevy:
Yah, this is fixing a bug that was introduced. The behavior isnt new. I think the minor bump is fine.

belevy:
Thank you for the fix :D

Issue #277, opened 2021-08-04
`distinctOnOrderBy` generates bad sql with nested exprs


NikitaRazmakhnin:
We noticed that when we use some nested expression with `distinctOnOrderBy` like `coalesce` generates bad sql in the beginning of a query:

```
let someVal = coalesce [ entity1?.Entity1Value, entity2^.Entity2Value ]
distinctOnOrderBy [ asc someVal ] $ do pure someVal
```
gives something like that
```
SELECT DISTINCT ON (COALESCE("entity1"."entity2_value",)
              COALESCE("entity1"."entity1_value", "entity2".entity2_value")
ORDER BY COALESCE("entity1"."entity1_value", "entity2".entity2_value")  ASC
```
So it omits second substitution-part of `COALESCE` expression for some reasons but the value expression itself works perfectly in `ORDER BY` part. So it behaves badly only in `DISTINCT ON (..)` clause.

Rewriting it into usual `distinctOn [don ...] $ do orderBy [...]` works perfectly.

PR #278, opened 2021-08-04
Fix distinctOnOrderBy with nested expression

NikitaRazmakhnin:
Should solve https://github.com/bitemyapp/esqueleto/issues/277.
Code assembling `DISTINCT ON` clause was stripping important part of important nested expressions. It worked well without nested expressions but with them it just cuts by spaces parts with arguments. So for example with `COALESCE(val, ?) ASC` it splits assembled expression into `["COALESCE(val,", "?)", "ASC"]` which causes bad sql for some cases.

I am not sure it is a good solution because I actually don't know is it possible to have other expressions here or not? Please, let me know if you see any other possible issues with that fix. :)

A related test-case is added.

After submitting your PR:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,9 @@
+3.5.2.2
+=======
+- @NikitaRazmakhnin
+  - [#278](https://github.com/bitemyapp/esqueleto/pull/278)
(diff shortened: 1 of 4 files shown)


parsonsmatt (approved):
fantastic, thanks!

Issue #231, opened 2020-11-27
Feature Request: Support for representing literal values as a Table


JonathanLorimer:
It would be amazing to have support for something like this:
```sql
WITH foo (col1, col2, col3) AS (
    SELECT * FROM (
        VALUES
            (1, 2, 3),
            (2, 3, 4)
        ) AS TMP
    )

SELECT * FROM foo
```

I don't know the esqueleto codebase very well, but it seems that it could be added to the `From` GADT like so:
```haskell
data From a where
    Values
        :: PersistEntity ent
        => [ent]
        -> From (SqlExpr (Entity ent))
    Table
        :: PersistEntity ent
        => From (SqlExpr (Entity ent))
    SubQuery
        :: ( SqlSelect a r
           , ToAlias a
           , ToAliasReference a
           )
        => SqlQuery a
        -> From a
    ...
```

belevy:
So this isn't a sql standard but rather postgresql specific as far as I can tell. I started work on a branch that eliminates the From GADT in favor of a typeclass. This would allow extensions of From to be made.in the backend specific modules.

JonathanLorimer:
Ahhhh, you are right, I only use postgresql so I just assumed other dbs had this capability. I would be happy to try and implement `Values` in the postgresql module once the migration to a typeclass has happened.

parsonsmatt:
I believe this is fixed by #284

PR #284, opened 2021-09-22
Add support of PostgreSQL-specific `VALUES(..)` expression

NikitaRazmakhnin:
At my company we really wanted to use `VALUES` Postgres expression to use it inside of `JOIN`
between large db data-set with some smaller user-input data-set without loading of data into app-memory.
`VALUES` allows us to do that and even gives performance benefit to our app.

I think it is a relevant issue https://github.com/bitemyapp/esqueleto/issues/231 :)

This PR tries to bring an ability to use like:

*SQL* : `SELECT * FROM (VALUES (a1), (a2) ...) as "vs"("v1")`
*Haskell*: `select $ from $ values [val 1, val 2, val 3]`

It also works for more than one columns as well (a bit doubting about aliases/refs assembling but works good so far). :)

I implemented it as part of `From` and thanks to latest refactoring from
GADTs to type-classes it was not so difficult :)

Few appropriate semi-integrational tests with other sql capabilities as join, sinigle/multicolumn were added as well.

Please, let me know if I made something in a wrong way - I am ready to try to fix :)

Thanks!

After submitting your PR:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,9 @@
+3.5.2.3
+=======
+- @NikitaRazmakhnin
+  - [#284](https://github.com/bitemyapp/esqueleto/pull/284)
+        - Add PostgreSQL-specific support of VALUES(..) literals
+
 3.5.2.2
 =======
 - @NikitaRazmakhnin
--- a/esqueleto.cabal
+++ b/esqueleto.cabal
@@ -1,7 +1,7 @@
 cabal-version: 1.12
 
 name:           esqueleto
-version:        3.5.2.2
+version:        3.5.2.3
 synopsis:       Type-safe EDSL for SQL queries on persistent backends.
 description:    @esqueleto@ is a bare bones, type-safe EDSL for SQL queries that works with unmodified @persistent@ SQL backends.  Its language closely resembles SQL, so you don't have to learn new concepts, just new syntax, and it's fairly easy to predict the generated SQL and optimize it for your backend. Most kinds of errors committed when writing SQL are caught as compile-time errors---although it is possible to write type-checked @esqueleto@ queries that fail at runtime.
                 .
--- a/src/Database/Esqueleto/PostgreSQL.hs
+++ b/src/Database/Esqueleto/PostgreSQL.hs
@@ -1,6 +1,8 @@
 {-# LANGUAGE CPP #-}
 {-# LANGUAGE FlexibleContexts #-}
+{-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE GADTs #-}
+{-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE OverloadedStrings #-}
 {-# LANGUAGE Rank2Types #-}
 {-# LANGUAGE ScopedTypeVariables #-}
@@ -26,6 +28,7 @@ module Database.Esqueleto.PostgreSQL
     , insertSelectWithConflict
     , insertSelectWithConflictCount
     , filterWhere
+    , values
     -- * Internal
     , unsafeSqlAggregateFunction
     ) where
@@ -39,9 +42,14 @@ import Control.Monad (void)
 import Control.Monad.IO.Class (MonadIO(..))
 import qualified Control.Monad.Trans.Reader as R
 import Data.Int (Int64)
+import qualified Data.List.NonEmpty as NE
+import Data.Maybe
 import Data.Proxy (Proxy(..))
 import qualified Data.Text.Internal.Builder as TLB
+import qualified Data.Text.Lazy as TL
 import Data.Time.Clock (UTCTime)
+import qualified Database.Esqueleto.Experimental as Ex
+import qualified Database.Esqueleto.Experimental.From as Ex
 import Database.Esqueleto.Internal.Internal hiding (random_)
 import Database.Esqueleto.Internal.PersistentImport hiding (upsert, upsertBy)
 import Database.Persist.Class (OnlyOneUniqueKey)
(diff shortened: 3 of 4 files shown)


src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -363,3 +371,75 @@ filterWhere aggExpr clauseExpr = ERaw noMeta $ \_ info ->
     in ( aggBuilder <> " FILTER (WHERE " <> clauseBuilder <> ")"
        , aggValues <> clauseValues
        )
+
+newtype PgValuesExprs a = PgValuesExprs { unPgValuesExprs :: NE.NonEmpty a }
```
parsonsmatt, line 375: Is this a Postgres specific feature?

[This article](https://modern-sql.com/feature/values) seems to indicate that it is a `SQL-92` feature, and should be supported by all SQL databases.

[MySQL supports `VALUES`](https://dev.mysql.com/doc/refman/8.0/en/values.html), as does [SQLite](https://www.sqlite.org/lang_select.html). So I think we can move this into `esqueleto` proper.

src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -363,3 +371,75 @@ filterWhere aggExpr clauseExpr = ERaw noMeta $ \_ info ->
...
+--            :| [ (val 20, val "twenty")
+--               , (val 30, val "thirty") ]
+--            )
+--      `InnerJoin` table User
+--      `on` (\((bound, _boundName) :& user) -> user^.UserAge >=. bound)
+--  groupBy bound
+--  pure (bound, count @Int $ user^.UserName)
+-- @
+--
+-- @since 3.5.2.3
```
parsonsmatt, line 443: 🙌🏻 this is dope

test/PostgreSQL/Test.hs
```diff
@@ -1270,6 +1271,53 @@ testLateralQuery = do
             let _ = res :: [(Entity Lord, Value (Maybe Int))]
             asserting noExceptions
 
+testValuesExpression :: SpecDb
+testValuesExpression = do
```
parsonsmatt, line 1275: blessed tests ❤️

parsonsmatt (approved):
This is awesome, thank you! 

Only thing is that I am *pretty* sure we can put this directly into `esqueleto` proper instead of `Esqueleto.Postgresql`. Can you try doing that and see if the tests all work out still?

parsonsmatt:
Try merging `master` once #285 is pulled in

src/Database/Esqueleto/PostgreSQL.hs
NikitaRazmakhnin: Ah sorry: my bad! Thanks for the links.

Sure, I can do that, but looks like other vendors requires different syntax, so I need to address that then :)

I will try to do that.

parsonsmatt: Oh, if they require different syntax, then let's not bother. The current solution is to keep anything that varies in these separate modules.

NikitaRazmakhnin: In some basic detail:
- MySql wants keyword `ROW(..)` for one row of values;
- SqLite requires some other placement of commas as far as I understand.. but worth to play a bit more.

I think consistency of API for vendors matters, so I am ready to try to cover other vendors as well.
But if you allow to go Postgres API earlier than other vendors - it would be awesome!

Otherwise - I could add necessary extensions for other vendors right here and ping you when it's ready :)

parsonsmatt: `esqueleto` doesn't currently do anything to differentiate the database provider - it's all `Sqlbackend` as far as we're concerned. So I'd rather not try and do anything clever here and just support PG directly. If someone else wants MySQL support or Sqlite, then we can add those in vendor-specific modules, and if we really want to unify it, we can try and do that.

NikitaRazmakhnin: Yep, understood: I also meant similar extensions in vendor-specific modules (like here) without unification (which probably also will present some code duplication but not a big deal I think).

src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -363,3 +371,75 @@ filterWhere aggExpr clauseExpr = ERaw noMeta $ \_ info ->
     in ( aggBuilder <> " FILTER (WHERE " <> clauseBuilder <> ")"
        , aggValues <> clauseValues
        )
+
+newtype PgValuesExprs a = PgValuesExprs { unPgValuesExprs :: NE.NonEmpty a }
+
+instance (ToSomeValues a, Ex.ToAliasReference a, Ex.ToAlias a) => Ex.ToFrom (PgValuesExprs a) a where
+  toFrom = fromValues . unPgValuesExprs
```
belevy, line 378: Is there a reason why you are going through `ToFrom` instead of just just making `values === fromValues`?

`ToFrom` is mostly there to support `SqlSetOperation` as a restricted subset and backwards compatibility with the `*Join` datatypes

src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -363,3 +371,75 @@ filterWhere aggExpr clauseExpr = ERaw noMeta $ \_ info ->
...
+-- share [mkPersist sqlSettings] [persistLowerCase|
+--   User
+--     name Text
+--     age Int
+--     deriving Eq Show
+--
+-- select $ do
+--  bound :& user <- from $
+--      values (   (val (10 :: Int), val ("ten" :: Text))
+--            :| [ (val 20, val "twenty")
```
belevy, line 434: I wish the OverloadedList extension wasn't such a foot cannon with NonEmpty and actually just gave a compilation error.

Nothing to change here, just me 😞 in haskell

src/Database/Esqueleto/PostgreSQL.hs
NikitaRazmakhnin: Ah, no reasons, this is just a lack of my knowledge of code-base and how it works in day when I wrote that :)
Thanks a lot for note, this is more clear for me now!

I have [fixed](https://github.com/bitemyapp/esqueleto/pull/284/commits/52373e5c430f5008ad4789528f519d60c9c2b39f) that :)

changelog.md
```diff
@@ -1,3 +1,9 @@
+3.5.2.3
```
parsonsmatt, line 1: Adding a new feature is a minor bump, so
```suggestion
3.5.3.0
```

esqueleto.cabal
```diff
@@ -1,7 +1,7 @@
 cabal-version: 1.12
 
 name:           esqueleto
-version:        3.5.2.2
+version:        3.5.2.3
```
```suggestion
version:        3.5.3.0
```

parsonsmatt (approved):
Adding functionality is a minor bump - otherwise this is great and looks good to go 😄 

I'll make the suggested modifications myself when I release for 3.5.3.0.

PR #301, opened 2022-03-05
Support upsert with empty updates

### LINKED ISSUE #300: Incorrect SQL query generated for PostgreSQL.upsert
I tried using [upsert](https://hackage.haskell.org/package/esqueleto-3.5.3.0/docs/Database-Esqueleto-PostgreSQL.html#v:upsert) function trying to produce `INSERT INTO ... VALUES ... ON CONFLICT DO NOTHING`. I passed an empty list of updates and got:
> SqlError {sqlState = "42601", sqlExecStatus = FatalError, sqlErrorMsg = "syntax error at or near \"WHERE\"", sqlErrorDetail = "", sqlErrorHint = ""}

I looked at logs and noticed that the following SQL statement is generated:
> INSERT INTO “foo_table”(“foo_id”,“foo_bar_id”,“foo_type”) VALUES ... ON CONFLICT (“foo_id”,“foo_bar_id”) DO UPDATE SET  WHERE “foo_table”.“foo_id” =... RETURNING ...`

As far as I understand, it is invalid because there is nothing between SET and WHERE. I think it would be more correct to produce `DO NOTHING` instead of `UPDATE SET WHERE...` in this case. Note that [insertSelectWithConflictCount](https://hackage.haskell.org/package/esqueleto-3.5.3.0/docs/src/Database.Esqueleto.PostgreSQL.html#insertSelectWithConflictCount) seems to handle this case correctly.

parsonsmatt:
Fixes #300 

After submitting your PR:


parsonsmatt:
Ah, dang. `upsert` is gonna need a new type signature.

In the event that we do an `INSERT ... ON CONFLICT DO NOTHING RETURNING ??`, `postgresql` *does not return anything* if we don't insert anything.

So the most proper type of `upsert` is then `upsert :: (PersistEntity ent, OnlyOneUniqueKey ent) => ent -> [Update ent] -> SqlPersistT m (Maybe (Entity ent))`

At least, as a default.

Playing around locally, the following form works:

```sql
WITH inserted AS (
    INSERT INTO "OneUnique" (name, value) 
    VALUES ("asdf", 0)
    ON CONFLICT (value)
    DO NOTHING
    RETURNING id, name, value
)
SELECT id, name, value
FROM inserted
UNION
SELECT id, name, value
FROM "OneUnique"
WHERE value = 0
```

But I have no idea what sorts of performance this has. [This StackOverflow question/answer](https://stackoverflow.com/questions/34708509/how-to-use-returning-with-on-conflict-in-postgresql) has some details and it looks to be somewhat complicated.

Maybe the type of `upsert` should really be `upsert :: entity -> NonEmpty (Update entity) -> SqlPersistT m (Entity entity)`. Then we have our no-op `insertOnConflictDoNothing :: entity -> SqlPersistT m (Maybe (Entity entity))`. Then you get:

```haskell
coolUpsert entity updates = 
    case NonEmpty.nonEmpty updates of
        Just nonEmptyUpdates ->
            Just <$> upsert entity nonEmptyUpdates
        Nothing ->
            insertOnConflictDoNothing entity
```

It might be tempting to write:

```haskel
insertOnConflictDoNothing :: record -> SqlPersistT m record 
insertOnCOnflictDoNothing rec = do
    xs <- rawSql thequery thevalues
    case xs of
        [] -> pure rec
        (a : _) -> pure a
```

But this only means that the uniqueness key is the same, not the rest of the values.

test/PostgreSQL/Test.hs
```diff
@@ -1038,18 +1066,27 @@ testInsertUniqueViolation =
...
+        u1e <- EP.upsert u1 [OneUniqueName =. val "fifth"]
+        liftIO $ entityVal u1e `shouldBe` u1
+        u2e <- EP.upsert u2 [OneUniqueName =. val "fifth"]
+        liftIO $ entityVal u2e `shouldBe` u2
+        u3e <- EP.upsert u3 [OneUniqueName =. val "fifth"]
+        liftIO $ entityVal u3e `shouldBe` u1{oneUniqueName="fifth"}
+    describe "With no updates" $ do
+        itDb "Works with no updates" $ do
+            _ <- EP.upsert u1 []
+            pure ()
```
parsonsmatt, line 1083: Well, this passes. Hooray.

test/PostgreSQL/Test.hs
```diff
@@ -1038,18 +1066,27 @@ testInsertUniqueViolation =
...
+        itDb "Works with no updates, twice" $ do
+            Entity u1Key u1' <- EP.upsert u1 []
+            Entity u1Key_ u1'' <- EP.upsert u1 { oneUniqueName = "Something Else" } []
+            pure ()
```
parsonsmatt, line 1087: Unfortunately, this fails with `ErrorCall "head: empty list"`.

The reason is that:

```sql
INSERT INTO "OneUnique" (name, value)
VALUES (?, ?)
ON CONFLICT DO NOTHING
RETURNING id, name, value
```

does not return anything if we don't perform an `INSERT` - so the return is `[]` which he `head` on and it goes boom.

`DO UPDATE` always does something. But we really don't want to do something silly like `DO UPDATE SET blah = EXCLUDED.blah` because this does unnecessary writes and triggers.

parsonsmatt (commented):
Dang, this is more complicated than I thought.

PR #315, opened 2022-05-24
Joseph/support update for locking

josephsumabat:

After submitting your PR:

### DIFF
--- a/test/PostgreSQL/Test.hs
+++ b/test/PostgreSQL/Test.hs
@@ -12,10 +12,11 @@
 module PostgreSQL.Test where
 
 import Control.Arrow ((&&&))
+import Control.Concurrent (forkIO)
 import Control.Monad (void, when)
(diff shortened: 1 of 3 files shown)


test/PostgreSQL/Test.hs
```diff
@@ -1227,6 +1231,192 @@ testCommonTableExpressions = do
             pure res
         asserting $ vals `shouldBe` fmap Value [2..11]
 
+testPostgresqlLocking :: SpecDb
+testPostgresqlLocking =
+    -- These tests use a create a fork in order to run two transactions concurrently.
+    -- This is necessary in order to test the locking behaviour
```
josephsumabat, line 1237: I could possibly abstract this forking behavior out into a function but it's a bit messy given the need for different queries and assertions. Please let me know if that would be strongly preferred though!

PR #316, opened 2022-05-24
Joseph/support for update of skip locked

josephsumabat:

After submitting your PR:

### DIFF
--- a/src/Database/Esqueleto/Internal/Internal.hs
+++ b/src/Database/Esqueleto/Internal/Internal.hs
@@ -1406,25 +1406,29 @@ data Insertion a
 -- are no guarantees that they will behave the same.
 --
 -- @since 2.2.7
-data LockingKind
-    = ForUpdate
+data LockingKind where
+    ForUpdate :: LockingKind
       -- ^ @FOR UPDATE@ syntax.  Supported by MySQL, Oracle and
       -- PostgreSQL.
       --
       -- @since 2.2.7
-    | ForUpdateSkipLocked
+    ForUpdateSkipLocked :: LockingKind
       -- ^ @FOR UPDATE SKIP LOCKED@ syntax.  Supported by MySQL, Oracle and
       -- PostgreSQL.
       --
       -- @since 2.2.7
-    | ForShare
+    ForShare :: LockingKind
       -- ^ @FOR SHARE@ syntax.  Supported by PostgreSQL.
       --
       -- @since 2.2.7
-    | LockInShareMode
+    LockInShareMode :: LockingKind
       -- ^ @LOCK IN SHARE MODE@ syntax.  Supported by MySQL.
       --
       -- @since 2.2.7
+    ForUpdateOfSkipLocked :: PersistEntity val => (SqlExpr (Entity val)) -> LockingKind
+      -- ^ @FOR UPDATE OF tablename SKIP LOCKED@ syntax. Supported by MySQL, and PostgreSQL
+      --
+      -- @since 3.5.4.0
 
 -- | Phantom class of data types that are treated as strings by the
 -- RDBMS.  It has no methods because it's only used to avoid type
@@ -2800,7 +2804,7 @@ toRawSql mode (conn, firstIdentState) query =
         , makeHaving     info havingClause
         , makeOrderBy    info orderByClauses
         , makeLimit      info limitClause
-        , makeLocking         lockingClause
+        , makeLocking    info lockingClause
         ]
 
 -- | Renders a 'SqlQuery' into a 'Text' value along with the list of
@@ -3033,13 +3037,15 @@ makeLimit (conn, _) (Limit ml mo) =
         v = maybe 0 fromIntegral
     in (TLB.fromText limitRaw, mempty)
 
-makeLocking :: LockingClause -> (TLB.Builder, [PersistValue])
-makeLocking = flip (,) [] . maybe mempty toTLB . Monoid.getLast
+makeLocking :: IdentInfo -> LockingClause -> (TLB.Builder, [PersistValue])
+makeLocking info = flip (,) [] . maybe mempty toTLB . Monoid.getLast
   where
     toTLB ForUpdate           = "\nFOR UPDATE"
     toTLB ForUpdateSkipLocked = "\nFOR UPDATE SKIP LOCKED"
     toTLB ForShare            = "\nFOR SHARE"
     toTLB LockInShareMode     = "\nLOCK IN SHARE MODE"
+    toTLB (ForUpdateOfSkipLocked (ERaw _ f))=
+      "\nFOR UPDATE OF " <> fst (f Never info) <> " SKIP LOCKED"
 
 parens :: TLB.Builder -> TLB.Builder
 parens b = "(" <> (b <> ")")
--- a/src/Database/Esqueleto/Internal/Internal.hs
+++ b/src/Database/Esqueleto/Internal/Internal.hs
@@ -3038,14 +3038,15 @@ makeLimit (conn, _) (Limit ml mo) =
     in (TLB.fromText limitRaw, mempty)
 
 makeLocking :: IdentInfo -> LockingClause -> (TLB.Builder, [PersistValue])
-makeLocking info = flip (,) [] . maybe mempty toTLB . Monoid.getLast
-  where
-    toTLB ForUpdate           = "\nFOR UPDATE"
-    toTLB ForUpdateSkipLocked = "\nFOR UPDATE SKIP LOCKED"
-    toTLB ForShare            = "\nFOR SHARE"
-    toTLB LockInShareMode     = "\nLOCK IN SHARE MODE"
-    toTLB (ForUpdateOfSkipLocked (ERaw _ f))=
-      "\nFOR UPDATE OF " <> fst (f Never info) <> " SKIP LOCKED"
+makeLocking info lockingClause = 
+  case Monoid.getLast lockingClause of
+    Just ForUpdate           -> ("\nFOR UPDATE", [])
+    Just ForUpdateSkipLocked -> ("\nFOR UPDATE SKIP LOCKED", [])
+    Just ForShare            -> ("\nFOR SHARE", [])
+    Just LockInShareMode     -> ("\nLOCK IN SHARE MODE", [])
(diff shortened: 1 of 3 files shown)


test/PostgreSQL/Test.hs
```diff
@@ -1227,6 +1231,192 @@ testCommonTableExpressions = do
             pure res
         asserting $ vals `shouldBe` fmap Value [2..11]
 
+testPostgresqlLocking :: SpecDb
+testPostgresqlLocking =
+    -- These tests use a create a fork in order to run two transactions concurrently.
```
josephsumabat, line 1236: I could possibly abstract this forking behavior out into a function but doing so feels a bit messy given the need for different queries and assertions. Please let me know if that would be strongly preferred though!

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1,18 +1,19 @@
 {-# LANGUAGE CPP #-}
-{-# LANGUAGE TypeApplications #-}
-{-# language DerivingStrategies, GeneralizedNewtypeDeriving #-}
 {-# LANGUAGE ConstraintKinds #-}
 {-# LANGUAGE DeriveDataTypeable #-}
+{-# LANGUAGE DerivingStrategies #-}
```
josephsumabat, line 4: This was changed from running `stylish-haskell` - Not sure if intended? notably running stylish-haskell naiively on this file runs into the error: `src/Database/Esqueleto/Internal/Internal.hs: RealSrcSpan SrcSpanPoint "src/Database/Esqueleto/Internal/Internal.hs" 412 1: lexical error in string/character literal at character 's'`. Removing the line and running it seems to allow it to run

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1426,11 +1426,15 @@ data LockingKind where
       -- ^ @LOCK IN SHARE MODE@ syntax.  Supported by MySQL.
       --
       -- @since 2.2.7
-    ForUpdateOfSkipLocked :: PersistEntity val => [(SqlExpr (Entity val))] -> LockingKind
+    ForUpdateOfSkipLocked :: [LockableEntity] -> LockingKind
       -- ^ @FOR UPDATE OF tablename SKIP LOCKED@ syntax. Supported by MySQL, and PostgreSQL
       --
       -- @since 3.5.4.2
 
+-- | Existential type to wrap table type entities for LockingKind
```
josephsumabat, line 1434: This is necessary since a `ForUpdateOfSkipLocked` can take a list of different types

test/PostgreSQL/Test.hs
parsonsmatt: Nah, I think this is fine - any generalization would have to take into account all kinds of weird behavior, so it's not clear the best way to abstract the behavior.

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -3033,13 +3051,17 @@ makeLimit (conn, _) (Limit ml mo) =
...
-makeLocking :: LockingClause -> (TLB.Builder, [PersistValue])
-makeLocking = flip (,) [] . maybe mempty toTLB . Monoid.getLast
-  where
-    toTLB ForUpdate           = "\nFOR UPDATE"
-    toTLB ForUpdateSkipLocked = "\nFOR UPDATE SKIP LOCKED"
-    toTLB ForShare            = "\nFOR SHARE"
-    toTLB LockInShareMode     = "\nLOCK IN SHARE MODE"
+makeLocking :: IdentInfo -> LockingClause -> (TLB.Builder, [PersistValue])
+makeLocking info lockingClause =
+  case Monoid.getLast lockingClause of
```
parsonsmatt, line 3056: This monoid behavior is disappointing. It renders composability of locking clauses pointless.

Since we're doing a major bump for this change, we can change the `Monoid` instance to select the *most restrictive* of two options.

I'm not sure about the proper way to do that, so it's fine to punt that to a separate PR and require it for a 3.6 release.

Then, if we have `ForUpdateOfSkipLocked rawNames`, we can `mappend` them rather than only taking the last, which would be bad.

parsonsmatt (changes requested):
Ok, a few changes to request:

1. `withAsync` refactor in the test suite
2. Version is `3.6.0.0` to account for breaking change (constructor for LockingKind was exposed)
3. Different API for use: `forUpdateOfSkipLocked :: ForUpdateEntry a => a -> SqlQuery ()`

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1406,25 +1407,42 @@ data Insertion a
...
+--
+-- @
+-- select $ do
+--   from $ \(p `InnerJoin` b) -> do
+--     on (p ^. PersonId ==. b ^. BlogPostAuthorId)
+--   locking (ForUpdateOfSkipLocked [LockableEntity p,LockableEntity b])
+--   return p
+-- @
+data LockableEntity where
+  LockableEntity :: PersistEntity val => (SqlExpr (Entity val)) -> LockableEntity
```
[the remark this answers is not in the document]
josephsumabat: I think we actually don't want `(SqlExpr (Maybe (Entity val))` to be available for the locking clause because of the error:

```
FOR UPDATE cannot be applied to the nullable side of an outer join
```
I'm not sure there's another case where we want to be able to do this with a nullable side?

Semantically `:&` is meant to represent "A left-precedence pair. Pronounced \"and\". Used to represent expressions
that have been joined together." so I'm a bit hesitant to use it for this purpose but maybe it's fine since the main use case is on joins?  I'm also not quite sure how to have an arbitrary number of tables with the tuple solution as well. 

For those reasons I do overall prefer a list since I think it lines up more closely with the syntax but don't feel super strongly about this so I'm happy to use this pattern instead if you strongly prefer it.

test/PostgreSQL/Test.hs
```diff
@@ -1227,6 +1231,201 @@ testCommonTableExpressions = do
...
+    -- This is necessary in order to test the locking behaviour
+    -- NOTE: If the forked process throws an exception then the test will hang as the assertion "MVar" never gets woken
```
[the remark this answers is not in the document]
josephsumabat: Fixed as part of: e25d95439d22a9324929caa75381b590a48e0ded!

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1406,25 +1407,42 @@ data Insertion a
...
       -- ^ @FOR SHARE@ syntax.  Supported by PostgreSQL.
       --
       -- @since 2.2.7
-    | LockInShareMode
+    LockInShareMode :: LockingKind
       -- ^ @LOCK IN SHARE MODE@ syntax.  Supported by MySQL.
       --
       -- @since 2.2.7
+    ForUpdateOfSkipLocked :: [LockableEntity] -> LockingKind
+      -- ^ @FOR UPDATE OF tablename SKIP LOCKED@ syntax. Supported by MySQL, and PostgreSQL
```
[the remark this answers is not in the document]
josephsumabat: Added `forUpdateOfSkipLocked`: 419d4e8c3bbd4df833fb475490dc5248a307a077!

I'm happy to submit a separate PR for the others (I would prefer to just have this one be for the `for update of` syntax). I can add them as well to this one if you strongly prefer that though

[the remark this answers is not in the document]
parsonsmatt: It's cool, I can make an issue and have the 3.6.0.0 release depend on that one

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1406,25 +1407,48 @@ data Insertion a
...
+--         `innerJoin` table @BlogPost
+--             `on` do
+--                 \(p :& bp) -> p ^. PersonId ==. b ^. BlogPostAuthorId
+--     forUpdateOfSkipLocked [LockableEntity p,LockableEntity b]
+--     return p
+-- @
+--
+-- @since 3.6.0.0
+data LockableEntity where
+  LockableEntity :: PersistEntity val => (SqlExpr (Entity val)) -> LockableEntity
```
parsonsmatt, line 1451: I think this will not make it into the released one - the idiom of using a type class argument that can accept tuples and `:&` is very nice syntactically. It also works on `SqlExpr (Maybe (Entity val))` which you get back from a `leftJoin`.

```haskell
class LockableEntity a where
    toLockableEntity :: a -> Whatever

instance LockableEntity (SqlExpr (Entity rec)) 

instance LockableEntity (SqlExpr (Maybe (Entity rec)))

instance (LockableEntity a, LocakbleEntity b) => LockableEntity (a :& b)

-- etc
```

These instances allow you to write, eg, 

```haskell
-- one table
t <- from $ table @Foo
forUpdateOfSkipLocked t

t :& s <- from $ table @Foo `leftJoin` table @Bar `on` ...
forUpdateOfSkipLocked s

all@(t :& s) <- from $ table @Foo `leftJoin` table @Bar `on` ...
forUpdateOfSkipLocked all
```

parsonsmatt (approved):
Awesome!

Going to approve now, but we'll want to wait to get the other 3.6.0.0 PRs and issues together before making a major release.

src/Database/Esqueleto/Internal/Internal.hs
josephsumabat: I think we actually don't want (SqlExpr (Maybe (Entity val)) to be available for the locking clause because of the error:

```
FOR UPDATE cannot be applied to the nullable side of an outer join
```

I'm not sure there's another case where we want to be able to do this with a nullable side since we would want the error to be caught by the type checker anyway. (ie we would only want to be able to do that for `inner joins` or the non nullable side of outer joins)

I don't feel super strongly about this though so happy to make the change if you prefer it (Just for clarification what is the `Whatever` type retyrned by `toLockableEntity`?)

parsonsmatt: That's a great catch!

In that case, we can even write a `TypeError` instance for `SqlExpr (Maybe (Entity rec))` 😄 

I'm happy to implement this after we get this merged. I think I will cut a branch for 3.5 soon and use that for backporting fixes and get 3.6 to an unreleased master.

PR #323, opened 2022-06-21
Add the ability to derive esqueleto records

9999years:
I recommend reviewing commits individually.

---

After submitting your PR:


esqueleto.cabal
```diff
@@ -2,7 +2,7 @@ cabal-version: 1.12
 
 name:           esqueleto
 
-version:        3.5.5.0
+version:        3.6.0.0
```
parsonsmatt, line 5: Since all changes here are an addition of a new module, then we can make this a minor change instead of a major change.

I don't *think* we've landed anything for 3.6 yet, so this can even get released now 😄 
```suggestion
version:        3.5.6.0
```

changelog.md
```diff
@@ -1,3 +1,8 @@
+3.6.6.0
```
```suggestion
3.5.6.0
```

changelog.md
```diff
@@ -1,3 +1,8 @@
...
+=======
+- @9999years
+    - #323
```
parsonsmatt, line 4:
```suggestion
    - [#323](https://github.com/bitemyapp/esqueleto/pull/323)
```

src/Database/Esqueleto/Record.hs
```diff
@@ -0,0 +1,423 @@
...
+        ++ constructorType
+
+    constructorName constructor =
+      case constructor of
+        (RecC name _) -> name
+        (NormalC name _fields) -> name
+        (InfixC _ty1 name _ty2) -> name
+        (ForallC _vars _constraints innerConstructor) -> constructorName innerConstructor
+        -- If there's GADTs where multiple constructors are declared with the
+        -- same type signature you're evil and furthermore this diagnostic will
```
parsonsmatt, line 420: :joy: 👍🏻

src/Database/Esqueleto/Record.hs
```diff
@@ -0,0 +1,423 @@
+{-# LANGUAGE ScopedTypeVariables #-}
+{-# LANGUAGE TemplateHaskellQuotes #-}
+
+module Database.Esqueleto.Record
+  ( deriveEsqueletoRecord
+  , takeColumns
```
parsonsmatt, line 6: `takeColumns` should be an internal implementation detail - we shouldn't need to export it.

src/Database/Esqueleto/Record.hs
```diff
@@ -0,0 +1,423 @@
...
+  --     fooName = fooName'
+  --
+  -- See `sqlSelectProcessRowPat` for the left-hand side of the patterns.
+  (statements, fieldExps) <-
+    unzip <$> forM (zip fields sqlFields) (\((fieldName', fieldType), (_, sqlType')) -> do
+      valueName <- newName (nameBase fieldName')
+      pure
+        ( BindS
+            (sqlSelectProcessRowPat fieldType valueName)
+            (AppTypeE (VarE 'takeColumns) sqlType')
```
parsonsmatt, line 305: So, you *would* need to export `takeColumns` (and then have it be imported at the use-sites of the esqueleto), *if* you wrote `mkName "takeColumns"` instead of `'takeColumns`. But this `'valueName` style gives you a globally qualified name that does not need to be imported in the file that uses the macro.

parsonsmatt: `mkName` gives you a dynamically bound variable, while `'valueName` gives you a name referring to the term `valueName` in scope at the *definition* of the macro.

Yesod actually does this when making the handlers - it does `mkName (verb ++ handlerName)`, and then *expects* you to import/define the relevant function so it can use that in the splice site.

src/Database/Esqueleto/Record.hs
```diff
@@ -0,0 +1,423 @@
...
+getRecordInfo :: Name -> Q RecordInfo
+getRecordInfo name = do
+  TyConI dec <- reify name
+  let (constraints, typeVarBinders, kind, constructors) =
+        case dec of
+          DataD constraints' _name typeVarBinders' kind' constructors' _derivingClauses ->
+            (constraints', typeVarBinders', kind', constructors')
+          NewtypeD constraints' _name typeVarBinders' kind' constructor' _derivingClauses ->
+            (constraints', typeVarBinders', kind', [constructor'])
+          _ -> error $ "Esqueleto records can only be derived for records and newtypes, but " ++ show name ++ " is neither"
```
parsonsmatt, line 146: Prefer `fail` to `error`, you'll get better error messages in `Q`

src/Database/Esqueleto/Record.hs
```diff
@@ -0,0 +1,423 @@
...
+      constructor = fromJust $ listToMaybe constructors
```
parsonsmatt, line 147: We may want to fail with a more sensible error here - believe it or not, you can write:

```haskell
data Void
```

which has no constructors.

```suggestion
    constructor <- 
      case constructors of
        (constr : _) -> pure constr
        [] -> fail "Cannot derive esqueleto record for a type with no constructors"
```

🤔

src/Database/Esqueleto/Record.hs
```diff
@@ -0,0 +1,423 @@
...
+-- | Transforms a record field type into a corresponding `SqlExpr` type.
+--
+-- * @'Entity' x@ is transformed into @'SqlExpr' ('Entity' x)@.
+-- * @'Maybe' ('Entity' x)@ is transformed into @'SqlExpr' ('Maybe' ('Entity' x))@.
+-- * @x@ is transformed into @'SqlExpr' ('Value' x)@.
+sqlFieldType :: Type -> Type
+sqlFieldType fieldType =
+  case fieldType of
+    -- Entity x -> SqlExpr (Entity x)
+    AppT (ConT ((==) ''Entity -> True)) _innerType -> AppT (ConT ''SqlExpr) fieldType
```
parsonsmatt, line 177: i really cannot get over how much i like this use of view patterns to match on a relatively complicated type!

parsonsmatt (commented):
Fantastic stuff so far! 

Let's get some tests written and verify this behaves like we want it to.

Simple smoke test is just "can we select this stuff out of the database?"

More complicated tests would involve `subSelect` and friends.

I'm also interested to see how this interacts with eg `CTE`s and subquery joins.

```haskell
data Foo = Foo { fooName :: String }

userToFooQuery = do
  u  <- from $ table @User
  pure Foo { fooName = u ^. UserName }

someJoin = select do
  (u :& f) <- from $ table @User
    `innerJoin` userToFooQuery
      `on` do
        \(u :& f) -> u ^. UserName ==. sqlFooName f
  pure (u, f)
````

File: test/Common/Record.hs [hunk 10]
```diff
@@ -0,0 +1,85 @@
...
+                          } -> True
+                 _ -> False)
+        liftIO $ sortedRecords !! 1
+          `shouldSatisfy`
+          (\case MyRecord { myName = "Some Guy"
+                          , myAge = Just 10
+                          , myUser = Entity _ User { userAddress  = Just addr1
+                                                   , userName = "Some Guy"
+                                                   }
+                          , myAddress = Just (Entity addr2 Address {addressAddress = "30-50 Feral Hogs Rd"})
```
parsonsmatt, line 83: 😂

parsonsmatt (approved):
I'll get the GHC 8.6 haddock failure figured out. This is awesome!

PR #324, opened 2022-07-07
Support nested records in `deriveEsqueletoRecord`

9999years:

After submitting your PR:

### DIFF
--- a/src/Database/Esqueleto/Record.hs
+++ b/src/Database/Esqueleto/Record.hs
@@ -5,8 +5,6 @@
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TemplateHaskellQuotes #-}
 {-# LANGUAGE TypeApplications #-}
-{-# LANGUAGE TypeFamilies #-}
-{-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE ViewPatterns #-}
 
 module Database.Esqueleto.Record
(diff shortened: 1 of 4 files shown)


src/Database/Esqueleto/Record.hs
```diff
@@ -379,22 +383,81 @@ sqlSelectProcessRowDec RecordInfo {..} = do
...
+          (\case InstanceD _overlap
+                           _constraints
+                           (extractSqlRecord originalType -> Just sqlRecord)
+                           _decs ->
+                             Just sqlRecord
+                 _ -> Nothing)
+
+  case filteredInstances of
+    sqlType : _ -> pure $ Just sqlType
+    _ -> pure Nothing
```
parsonsmatt, line 459:
```suggestion
  pure $ listToMaybe filteredInstances
```

src/Database/Esqueleto/Record.hs
```diff
@@ -379,22 +383,81 @@ sqlSelectProcessRowDec RecordInfo {..} = do
...
+        case instanceTy of
+          (ConT ((==) ''SqlSelect -> True))
+            `AppT` sqlTy
+            `AppT` ((==) originalTy -> True) -> Just sqlTy
+          _ -> Nothing
+
+      -- Filter `instances` to the instances which match `originalType`.
+      filteredInstances :: [Type]
+      filteredInstances =
+        flip mapMaybe instances
```
parsonsmatt, line 449: hm, `listToMaybe . mapMaybe f` is also `find (isJust . f)` - may be easier or more concise to write like that?

parsonsmatt (approved):
this looks great! just needs a few tests written to ensure it works

parsonsmatt (approved):

src/Database/Esqueleto/Record.hs
9999years: I don't think so, because `find` only returns the element but I also need the `mapMaybe` part to transform and extract the sql type.

parsonsmatt: Oh, good point!

parsonsmatt:
released as 3.5.6.1

PR #326, opened 2022-08-01
Skip Haddock for GHC 8.6.5

parsonsmatt:
[GHC 8.6.5 is broken with module re-exports](https://github.com/haskell/haddock/issues/979#issuecomment-485341270). We can either re-export everything, or just skip building Haddock. I'd rather skip building Haddocks for an old GHC than make the code experience worse, or more fragile.

After submitting your PR:

### DIFF
--- a/.github/workflows/haskell.yml
+++ b/.github/workflows/haskell.yml
@@ -72,5 +72,6 @@ jobs:
             ${{ runner.os }}-${{ matrix.ghc }}-
       - run: cabal v2-build --disable-optimization -j $CONFIG
       - run: cabal v2-test --disable-optimization -j $CONFIG --test-options "--fail-on-focus"
-      - run: cabal v2-haddock -j $CONFIG
+      - if: ${{ matrix.ghc != "8.6.5" }}
+        run: cabal v2-haddock -j $CONFIG
       - run: cabal v2-sdist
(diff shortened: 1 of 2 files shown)


parsonsmatt:
Here's the error we get:

```
esqueleto            >   Missing documentation for:
esqueleto            >     ExprParser (src/Database/Esqueleto/Internal/ExprParser.hs:53)
esqueleto            >     onExpr (src/Database/Esqueleto/Internal/ExprParser.hs:55)
esqueleto            >     skipToEscape (src/Database/Esqueleto/Internal/ExprParser.hs:62)
esqueleto            >     parseEscapedIdentifier (src/Database/Esqueleto/Internal/ExprParser.hs:65)
esqueleto            >     parseTableAccess (src/Database/Esqueleto/Internal/ExprParser.hs:72)
esqueleto            >     parseEscapedChars (src/Database/Esqueleto/Internal/ExprParser.hs:79)
esqueleto            > haddock: internal error: internal: extractDecl
esqueleto            > CallStack (from HasCallStack):
esqueleto            >   error, called at utils/haddock/haddock-api/src/Haddock/Interface/Create.hs:1116:12 in main:Haddock.Interface.Create
```

It's a Haddock bug that we're somehow getting triggered.

Seems related to [this issue](https://github.com/haskell/haddock/issues/979), which indicates that Haddock is just broken here for GHC 8.6 (see [this comment](https://github.com/haskell/haddock/issues/979#issuecomment-485341270). The fix is dropping the module re-export and replacing with an explicit export list.

But GHC 8.6.5 itself isn't hard to support, so I'm not terribly worried about keeping it around. If it does become a trouble, then we'll drop it. I think it makes sense to retain support for GHC 8.10.7 until people aren't using it much, which will be a *while*.

parsonsmatt:
if anyone is seeing this and wondering *why*, i'd be happy to accept a PR that just fixes it and re-enables, fwiw

PR #328, opened 2022-08-15
Add `ToAlias`, `ToAliasReference` for <=16-tuples

ivanbakel:
This adds a `ToAlias` and `ToAliasReference` instance for every tuple size from 9 to 16. These tuple sizes are supported elsewhere, but those typeclasses only had instances for tuples up to size 8.

This also adds `from*` functions for tuples of size 9-16, which didn't exist before.

Do I need to add Haddocks and `@since` declarations for these functions, given that they're in the `Internal` module?

After submitting your PR:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,10 @@
+3.5.7.0
+=======
+- @ivanbakel
+    - [#328](https://github.com/bitemyapp/esqueleto/pull/328)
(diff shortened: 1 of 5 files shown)


parsonsmatt (approved):
This is great, thanks!

I'm going to merge and release for 3.5.10.0

PR #331, opened 2022-08-19
Add customisation of auto-gen'ed SQL record names

ivanbakel:
This adds an Aeson-inspired ability to configure how `deriveEsqueletoRecord` generates the SQL record type, in particular which names it generates for the record type itself and its fields.

This is because `DuplicateRecordFields` is a painful extension to use, and avoiding it through module structuring is also painful. A nicer solution for avoiding duplicate record fields is to simply not generate them, and allow the user to specify how record fields should be generated so they don't overlap.

After submitting your PR:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,16 @@
+3.5.8.0
+=======
+- @ivanbakel
+    - [#331](https://github.com/bitemyapp/esqueleto/pull/331)
+        - Add `deriveEsqueletoRecordWith` to derive Esqueleto instances for
+          records using custom deriving settings.
+        - Add `DeriveEsqueletoRecordSettings` to control how Esqueleto record
+          instances are derived.
+        - Add `sqlNameModifier` to control how Esqueleto record instance
+          deriving generates the SQL record type name.
+        - Add `sqlFieldModifier` to control how Esqueleto record instance
+          deriving generates the SQL record fields.
+
 3.5.7.1
 =======
 - @belevy
--- a/src/Database/Esqueleto/Record.hs
+++ b/src/Database/Esqueleto/Record.hs
@@ -9,6 +9,10 @@
 
 module Database.Esqueleto.Record
   ( deriveEsqueletoRecord
+  , deriveEsqueletoRecordWith
+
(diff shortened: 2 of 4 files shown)


src/Database/Esqueleto/Record.hs
```diff
@@ -115,8 +119,51 @@ import Data.Maybe (mapMaybe, fromMaybe, listToMaybe)
...
 -- @since 3.5.6.0
 deriveEsqueletoRecord :: Name -> Q [Dec]
-deriveEsqueletoRecord originalName = do
-  info <- getRecordInfo originalName
+deriveEsqueletoRecord = deriveEsqueletoRecordWith defaultSettings
+
+-- | Codegen settings for 'deriveEsqueletoRecordWith'.
+--
+-- @since 3.5.8.0
+data Settings = Settings
```
parsonsmatt, line 127: The `Settings` name is going to be a really common conflict. Let's give it a more specific name:

```suggestion
data DeriveEsqueletoRecordSettings = DeriveEsqueletoRecordSettings
```

parsonsmatt (changes requested):
This is great, thank you!

Only one comment on the name for `Settings` - let's make something less ambiguously named

src/Database/Esqueleto/Record.hs
ivanbakel: If you prefer. I tend to think of the module name as disambiguating these kinds of general names.

changelog.md
```diff
@@ -1,3 +1,16 @@
+3.5.8.0
+=======
+- @ivanbakel
+    - [#331](https://github.com/bitemyapp/esqueleto/pull/331)
+        - Add `deriveEsqueletoRecordWith` to derive Esqueleto instances for
+          records using custom deriving settings.
+        - Add `Settings` to control how Esqueleto record instances are
```
jappeace, line 7:
```suggestion
        - Add `DeriveEsqueletoRecordSettings` to control how Esqueleto record instances are
```

parsonsmatt (approved):
awesome, thanks so much!

PR #334, opened 2022-08-23
Add test and fix for issue 299

### LINKED ISSUE #299: Ambiguous column reference when using union result in subquery/CTE with persistent-postgresql
@mauriciofierrom posted [this issue in `persistent`](https://github.com/yesodweb/persistent/issues/1363), but I think it belongs here.

Copying:

> ### Description
> With the `persistent-postgresql` library, when using the result of a `union_` in a subquery/CTE, the generated SQL repeats the "v" alias, resulting in an error about an ambiguous column reference. The generated SQL for the `persistent-sqlite` backend worked fine.
> 
> ### Envrironment
> Here's the [gist](https://gist.github.com/mauriciofierrom/da37b32c54aa4a6a3e67c588304b4d8e) with the result of running the relevant commands listed in the issue template.
> 
> ### Script
> Here's the [script](https://gist.github.com/mauriciofierrom/6803e44e5354b04b7be89c56707df067) that results in the exception.
> 
> If considered relevant I'd like to work on this and if that would be OK, any guidance would be highly appreciated.

belevy:
@parsonsmatt This should fix #299 by aggressively re-aliasing. Previously, if an already aliased value was discovered it would not be realiased. This caused an issue where an alias from a lower level could be mixed with an alias from a higher level. This only showed up as an issue when using set operations due to its "smarts" that try to avoid using more Idents than absolutely necessary. 

I guess this is a minor version bump?

I would like your eyes on this before I merge.


parsonsmatt (approved):
Looks good to me! Thank you 🎉

parsonsmatt:
As for version - there shouldn't be any changes to the behavior of the library, up to alpha equivalence on SQL queries. And it's a bug fix, not a new feature. I'd probably do a patch release.

PR #336, opened 2022-08-31
Revert change to ToAlias and refix union bug

belevy:
#334  broke the behavior of nested subqueries

`from $ from $ from table` would not correctly re-alias the values.

This PR reverts that fix and substitutes it with a different one that addresses union(set operations) directly by changing how it reuses variable names. 

Lesson Learned: Do not try to reuse variable names unless you want pain

### DIFF
--- a/test/PostgreSQL/Test.hs
+++ b/test/PostgreSQL/Test.hs
@@ -1319,7 +1319,7 @@ testValuesExpression = do
                                        , (Value 3, Value "str3", Value Nothing) ]
 
 testSubselectAliasingBehavior :: SpecDb
-testSubselectAliasingBehavior = focus $ do
+testSubselectAliasingBehavior = do
     describe "Aliasing behavior" $ do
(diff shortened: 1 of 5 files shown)


test/PostgreSQL/Test.hs
```diff
@@ -1318,9 +1318,14 @@ testValuesExpression = do
                                        , (Value 2, Value "str2", Value $ Just 2.5)
                                        , (Value 3, Value "str3", Value Nothing) ]
 
-testSubselectUnionError :: SpecDb
-testSubselectUnionError = do
-    describe "Subselect union error" $ do
+testSubselectAliasingBehavior :: SpecDb
+testSubselectAliasingBehavior = focus $ do
```
parsonsmatt, line 1322:
```suggestion
testSubselectAliasingBehavior = do
```

parsonsmatt:
released as 3.5.8.1

PR #339, opened 2022-10-17
Support more locking kind

josephsumabat:
**Context:**
Support `FOR UPDATE Of tablename` and `FOR SHARE OF tablename` syntax for Postgresql:

This allows a the `FOR UPDATE SKIP LOCKED`  behaviour when you have a `LEFT OUTER JOIN` with a nullable side.  Currently you will get the error:
```
FOR UPDATE cannot be applied to the nullable side of an outer join
```
The workaround currently using esqueleto is to do seperate queries and then join them manually if you want to select rows for an update while skipping locked rows (or to just do a raw query). I ran into this situation at work.

See: https://www.postgresql.org/docs/current/sql-select.html
namely:
```
FOR { UPDATE | NO KEY UPDATE | SHARE | KEY SHARE } [ OF table_name [, ...] ] [ NOWAIT | SKIP LOCKED ] [...] ]
```
also documented in MySQL: 
https://dev.mysql.com/doc/refman/8.0/en/select.html

```
    [FOR {UPDATE | SHARE}
        [OF tbl_name [, tbl_name] ...]
        [NOWAIT | SKIP LOCKED]
      | LOCK IN SHARE MODE]
    [into_option]
```
Usage example would be
```
select $ do
    (p :& bp) <- from $ 
        table @Person
        `innerJoin` table @BlogPost
            `on` do
                \(p :& b) -> p ^. PersonId ==. b ^. BlogPostAuthorId
                EP.forUpdateOf (p :& b) EP.skipLocked
    return p

```

After submitting your PR:

### DIFF
--- a/src/Database/Esqueleto/Experimental/From/Join.hs
+++ b/src/Database/Esqueleto/Experimental/From/Join.hs
@@ -27,21 +27,6 @@ import Database.Esqueleto.Internal.Internal hiding
        (From(..), from, fromJoin, on)
 import GHC.TypeLits
 
--- | A left-precedence pair. Pronounced \"and\". Used to represent expressions
--- that have been joined together.
---
--- The precedence behavior can be demonstrated by:
---
--- @
--- a :& b :& c == ((a :& b) :& c)
--- @
---
--- See the examples at the beginning of this module to see how this
--- operator is used in 'JOIN' operations.
-data (:&) a b = a :& b
-    deriving (Eq, Show)
-infixl 2 :&
-
 instance (ToMaybe a, ToMaybe b) => ToMaybe (a :& b) where
     type ToMaybeT (a :& b) = (ToMaybeT a :& ToMaybeT b)
     toMaybe (a :& b) = (toMaybe a :& toMaybe b)
--- a/src/Database/Esqueleto/Experimental/From/Join.hs
+++ b/src/Database/Esqueleto/Experimental/From/Join.hs
@@ -13,6 +13,9 @@
 {-# OPTIONS_GHC -fno-warn-orphans #-}
 
 module Database.Esqueleto.Experimental.From.Join
+  (module Database.Esqueleto.Experimental.From.Join
+  , (:&)
+  )
     where
 
 import Data.Bifunctor (first)
--- a/src/Database/Esqueleto/Internal/Internal.hs
+++ b/src/Database/Esqueleto/Internal/Internal.hs
@@ -75,6 +75,7 @@ import Database.Persist.Sql.Util
 import Database.Persist.SqlBackend
 import GHC.Records
 import Text.Blaze.Html (Html)
+import qualified Data.List as List
 
 -- | (Internal) Start a 'from' query with an entity. 'from'
 -- does two kinds of magic using 'fromStart', 'fromJoin' and
@@ -1449,37 +1450,6 @@ data LockingKind
       --
       -- @since 2.2.7
 
-newtype (ComparableLockStrength a) => StrongestLock a =
-  StrongestLock { getStrongestLock :: Maybe a}
-
-class ComparableLockStrength a where
-  -- | Create a lock that is at least as strong as both locks and at most as
-  -- strong as the stronger lock
-  -- Should satisfy:
-  --    If compare a b = GT
-  --        compareLockStrength (unionLocks a b) b = GT
-  --        compareLockStrength a (unionLocks a b) = EQ
-  --    If compare a b = EQ
-  --        compareLockStrength (unionLocks a b) b = EQ
-  --        compareLockStrength a (unionLocks a b) = EQ
-  --    If compare a b = LT
-  --        compareLockStrength (unionLocks a b) b = EQ
-  --        compareLockStrength a (unionLocks a b) = LT
-  unionLocks :: ComparableLockStrength a => a -> a -> a
-  compareLockStrength :: ComparableLockStrength a => a -> a -> Ordering
-
-instance ComparableLockStrength a => Semigroup (StrongestLock a) where
-  (<>) (StrongestLock Nothing) pright = pright
-  (<>) pleft (StrongestLock Nothing) = pleft
-  (<>) (StrongestLock (Just pright)) (StrongestLock (Just pleft)) =
-    case compareLockStrength pleft pright of
-      GT -> StrongestLock (Just pright)
-      EQ -> StrongestLock (Just $ unionLocks pright pleft)
-      LT -> StrongestLock (Just pleft)
-
-instance ComparableLockStrength a => Monoid (StrongestLock a) where
-  mempty = StrongestLock Nothing
-
 -- | Postgres specific locking, used only internally
 --
 -- @since 3.5.9.0
@@ -1491,18 +1461,6 @@ data PostgresLockingKind =
     , postgresOnLockedBehavior :: OnLockedBehavior
     }
 
-instance ComparableLockStrength PostgresLockingKind where
-  unionLocks pl1 pl2 =
-    let newOfClause = do
-          (Of e1) <- postgresLockingOfClause pl1
-          (Of e2) <- postgresLockingOfClause pl2
-          pure $ Of (e1 :& e2) in
-    case compareLockStrength pl1 pl2 of
-      GT -> pl1 { postgresLockingOfClause = newOfClause }
-      EQ -> pl1 { postgresLockingOfClause = newOfClause }
-      LT -> pl2 { postgresLockingOfClause = newOfClause }
-  compareLockStrength a b = compare (postgresRowLevelLockStrength a) (postgresRowLevelLockStrength b)
-
 -- Arranged in order of lock strength
 data PostgresRowLevelLockStrength =
     PostgresForUpdate
@@ -1510,10 +1468,7 @@ data PostgresRowLevelLockStrength =
   deriving (Ord, Eq)
 
 data LockingOfClause where
-  Of :: LockableEntity a => a -> LockingOfClause
-
-unionOfClause :: LockingOfClause -> LockingOfClause -> LockingOfClause
-unionOfClause (Of a) (Of b) = (Of (a :& b))
+  LockingOfClause :: LockableEntity a => a -> LockingOfClause
 
 data OnLockedBehavior =
   NoWait
@@ -2127,30 +2082,21 @@ instance Monoid LimitClause where
 data LockingClause =
   GeneralLockingClause LockingKind
   -- ^ Locking clause not specific to any database implementation
-  | PostgresLockingClause PostgresLockingKind
+  | PostgresLockingClauses [PostgresLockingKind]
   -- ^ Locking clause specific to postgres
   | NoLockingClause
 
--- | Monoid instance prioritization is:
--- 1. "Strongest" lock (FOR UPDATE > FOR SHARE) if comparing two postgres
--- specific locks of differing strength
--- 2. Taking a "union" of postgres locks of equal strength (two locks which lock
--- multiple tables will lock tables specified in either lock) for lock strength. Lock BEHAVIOUR will prioritize the first lock
--- 3. Rightmost (last) locking clause if between two general locks for
--- backwards compatibility
--- 4. ForUpdate/ForUpdateSkipLocked > Postgres For Update Of > ForShare/LockInShareMode > PostgresFor Share Of if comparing between postgres and non-postgres locks
 instance Semigroup LockingClause where
-  (<>) (PostgresLockingClause pleft) (PostgresLockingClause pright) =
-    Maybe.maybe mempty PostgresLockingClause
-    (getStrongestLock $ StrongestLock (Just pleft) <> StrongestLock (Just pright))
+  -- Postgres allows us to have multiple locking clauses
+  (<>) (PostgresLockingClauses pleft) (PostgresLockingClauses pright) = PostgresLockingClauses (pleft <> pright)
   -- For backwards compatibility reasons we use the monoid instance of last for
   -- general locking cluase
   (<>) (GeneralLockingClause gleft) (GeneralLockingClause gright)  =
     Maybe.maybe mempty GeneralLockingClause
       (Monoid.getLast $ Monoid.Last (Just gleft) <> Monoid.Last (Just gright))
   
-  (<>) gleft@(GeneralLockingClause _) (PostgresLockingClause _)  = gleft
-  (<>) (PostgresLockingClause _) gright@(GeneralLockingClause _) = gright
+  (<>) gleft@(GeneralLockingClause _) (PostgresLockingClauses _)  = gleft
+  (<>) (PostgresLockingClauses _) gright@(GeneralLockingClause _) = gright
   (<>) mleft NoLockingClause = mleft
   (<>) NoLockingClause mright = mright
 --
@@ -3230,13 +3176,19 @@ makeLocking _ (GeneralLockingClause lockingClause) =
     ForUpdateSkipLocked -> ("\nFOR UPDATE SKIP LOCKED", [])
     ForShare            -> ("\nFOR SHARE", [])
     LockInShareMode     -> ("\nLOCK IN SHARE MODE", [])
-makeLocking info (PostgresLockingClause l) =
-    makeLockingStrength (postgresRowLevelLockStrength l)
-      <> plain " "
-      <> makeOfClause (postgresLockingOfClause l)
-      <> plain " "
-      <> makeLockingBehavior (postgresOnLockedBehavior l)
+makeLocking info (PostgresLockingClauses clauses) =
+  List.foldl' combineBuilderValPairs ("",[]) (makePostgresLockingClauses <$> clauses)
     where
+      combineBuilderValPairs (builder1, persistvals1) (builder2,persistvals2) =
+        (builder1 <> builder2 <> "\n", persistvals1 <> persistvals2)
+
+      makePostgresLockingClauses :: PostgresLockingKind -> (TLB.Builder , [PersistValue])
+      makePostgresLockingClauses l =
+        makeLockingStrength (postgresRowLevelLockStrength l)
+          <> plain " "
+          <> makeOfClause (postgresLockingOfClause l)
+          <> plain " "
+          <> makeLockingBehavior (postgresOnLockedBehavior l)
       makeLockingStrength :: PostgresRowLevelLockStrength -> (TLB.Builder, [PersistValue])
       makeLockingStrength PostgresForUpdate = plain "FOR UPDATE"
       makeLockingStrength PostgresForShare = plain "FOR SHARE"
@@ -3247,7 +3199,7 @@ makeLocking info (PostgresLockingClause l) =
       makeLockingBehavior Wait = plain ""
 
       makeOfClause :: Maybe LockingOfClause -> (TLB.Builder, [PersistValue])
-      makeOfClause (Just (Of lockableEnts)) = plain "OF " <> makeLockableEntity info lockableEnts
+      makeOfClause (Just (LockingOfClause lockableEnts)) = plain "OF " <> makeLockableEntity info lockableEnts
       makeOfClause Nothing = plain ""
 
       plain v = (v,[])
--- a/src/Database/Esqueleto/Internal/Internal.hs
+++ b/src/Database/Esqueleto/Internal/Internal.hs
@@ -14,8 +14,8 @@
 {-# LANGUAGE Rank2Types #-}
 {-# LANGUAGE ScopedTypeVariables #-}
 {-# LANGUAGE TypeApplications #-}
-{-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE TypeFamilies #-}
+{-# LANGUAGE TypeOperators #-}
 {-# LANGUAGE UndecidableInstances #-}
 
 {-# OPTIONS_GHC -fno-warn-redundant-constraints #-}
@@ -396,7 +396,7 @@ rand = ERaw noMeta $ \_ _ -> ("RANDOM()", [])
 having :: SqlExpr (Value Bool) -> SqlQuery ()
 having expr = Q $ W.tell mempty { sdHavingClause = Where expr }
 
--- | Add a locking clause to the query.  Please read
+-- | Add a general locking clause to the query.  Please read
 -- 'LockingKind' documentation and your RDBMS manual.
 --
 -- If multiple calls to 'locking' are made on the same query,
@@ -406,6 +406,9 @@ having expr = Q $ W.tell mempty { sdHavingClause = Where expr }
 locking :: LockingKind -> SqlQuery ()
 locking kind = putLocking $ GeneralLockingClause kind
 
+-- | Helper to add a any type of locking clause to a query
+--
+-- @since 3.5.9.0
 putLocking :: LockingClause -> SqlQuery ()
 putLocking clause = Q $ W.tell mempty { sdLockingClause = clause }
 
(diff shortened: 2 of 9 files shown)


src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1396,6 +1402,21 @@ data Update
 -- | Phantom type used by 'insertSelect'.
 data Insertion a
 
+-- | A left-precedence pair. Pronounced \"and\". Used to represent expressions
```
josephsumabat, line 1405: Moved here to access for locking kind

src/Database/Esqueleto/Experimental/From/Join.hs
```diff
@@ -27,21 +27,6 @@ import Database.Esqueleto.Internal.Internal hiding
...
---
--- @
--- a :& b :& c == ((a :& b) :& c)
--- @
---
--- See the examples at the beginning of this module to see how this
--- operator is used in 'JOIN' operations.
-data (:&) a b = a :& b
-    deriving (Eq, Show)
-infixl 2 :&
```
parsonsmatt, line 43: This is a public module, so removing this declaration without re-exporting it is a breaking change.

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1405,26 +1426,105 @@ data Insertion a
...
       --
       -- @since 2.2.7
 
+-- | Postgres specific locking, used only internally
+--
+-- @since 3.5.8.0
+data PostgresLockingKind where
+  PostgresLockingKind :: PostgresRowLevelLockStrength -> OnLockedBehavior -> PostgresLockingKind
+
+instance Eq PostgresLockingKind where
```
josephsumabat, line 1455: This is a bit weird since it's comparing equality of locking strength rather than equalilty itself - Maybe this should all be done in a separate function instead of using `Ord`?

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -401,7 +404,10 @@ having expr = Q $ W.tell mempty { sdHavingClause = Where expr }
 --
 -- @since 2.2.7
 locking :: LockingKind -> SqlQuery ()
-locking kind = Q $ W.tell mempty { sdLockingClause = Monoid.Last (Just kind) }
+locking kind = putLocking $ GeneralLockingClause kind
+
+putLocking :: LockingClause -> SqlQuery ()
+putLocking clause = Q $ W.tell mempty { sdLockingClause = clause }
```
parsonsmatt, line 410: Please add a doc comment 

```suggestion
-- |
--
-- @since {this version}
putLocking :: LockingClause -> SqlQuery ()
putLocking clause = Q $ W.tell mempty { sdLockingClause = clause }
```

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1983,10 +2083,30 @@ instance Semigroup LimitClause where
...
+  -- ^ Locking clause not specific to any database implementation
+  | PostgresLockingClause PostgresLockingKind
+  -- ^ Locking clause specific to postgres
+  | NoLockingClause
+
+-- | Monoid instance prioritization is:
+-- 1. "Strongest" lock (FOR UPDATE > FOR SHARE) if comparing two postgres
+-- specific locks of differing strength
+-- 2. Rightmost (last) locking clause otherwise
+instance Semigroup LockingClause where
```
josephsumabat, line 2099: This should satisfy the associativity and identity laws

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1405,26 +1426,105 @@ data Insertion a
...
+    LockInShareMode :: LockingKind
       -- ^ @LOCK IN SHARE MODE@ syntax.  Supported by MySQL.
       --
       -- @since 2.2.7
 
+-- | Postgres specific locking, used only internally
+--
+-- @since 3.5.8.0
+data PostgresLockingKind where
+  PostgresLockingKind :: PostgresRowLevelLockStrength -> OnLockedBehavior -> PostgresLockingKind
```
parsonsmatt, line 1453: GADT syntax is a bit of a red herring here - mind using ordinary data constructor syntax?

```suggestion
data PostgresLockingKind 
    = PostgresLockingKind PostgresRowLevelLockStrength OnLockedBehavior PostgresLockingKind
```

A record may be a good idea, too, just so we have keyword access

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1405,26 +1426,105 @@ data Insertion a
...
+
+instance Eq PostgresLockingKind where
+  (==) a b = compare a b == EQ
+
+instance Ord PostgresLockingKind where
+  compare (PostgresLockingKind rowstr1 onLockedBehavior1) (PostgresLockingKind rowstr2 onLockedBehavior2) =
+    case compare rowstr1 rowstr2 of
+      EQ -> compare onLockedBehavior1 onLockedBehavior2
+      rowStrCmp -> rowStrCmp
```
parsonsmatt, line 1462: This is equivalent to the stock derived `Ord` instance

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1405,26 +1426,105 @@ data Insertion a
...
+
+data PostgresRowLevelLockStrength =
+    PostgresForUpdate (Maybe LockingOfClause)
+    | PostgresForShare (Maybe LockingOfClause)
```
parsonsmatt, line 1466: Can we hoist this out into the `Database.Esqueleto.Postgres` module?

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1405,26 +1426,105 @@ data Insertion a
...
+
+-- | Eq instance of RowLevelLockStrength refers to the equality of the 
+-- strength of lock specifically
+instance Eq PostgresRowLevelLockStrength where
+  (==) a b = compare a b == EQ
```
parsonsmatt, line 1471: I think this is a dubious instance. You can easily construct values where two dissimilar things are considered `==` by this. How is this instance being used? Would it be better for this behavior to be in a separate function?

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1405,26 +1426,105 @@ data Insertion a
...
+
+-- | Ord instance of RowLevelLockStrength refers to the order of the strength
+-- of lock only. 
+instance Ord PostgresRowLevelLockStrength where
+  compare (PostgresForUpdate _) (PostgresForUpdate _) = EQ
+  compare (PostgresForUpdate _) _  = GT
+  compare _ (PostgresForUpdate _)  = LT
+  compare (PostgresForShare _) (PostgresForShare _)  = EQ
```
parsonsmatt, line 1479: this does also feel dubious to me

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1405,26 +1426,105 @@ data Insertion a
...
+--             `on` do
+--                 \(p :& bp) -> p ^. PersonId ==. b ^. BlogPostAuthorId
+--     forUpdateOf (p :& b) skipLocked
+--     return p
+-- @
+class LockableEntity a where
+  flattenLockableEntity :: a -> [LockableSqlExpr]
+  makeLockableEntity :: LockableEntity a => IdentInfo -> a -> (TLB.Builder, [PersistValue])
+  makeLockableEntity info lockableEntity =
+    uncommas' $ (\(LockableSqlExpr (ERaw _ f)) -> f Never info) <$> flattenLockableEntity lockableEntity
```
parsonsmatt, line 1517: If we're not overriding this in any instances, then it can be supplied as a stand-alone function instead of a class member instance. Haven't seen all instances yet here so this may be a redundant comment

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1405,26 +1426,105 @@ data Insertion a
...
+--     (p :& bp) <- from $ 
+--         table @Person
+--         `innerJoin` table @BlogPost
+--             `on` do
+--                 \(p :& bp) -> p ^. PersonId ==. b ^. BlogPostAuthorId
+--     forUpdateOf (p :& b) skipLocked
+--     return p
+-- @
+class LockableEntity a where
+  flattenLockableEntity :: a -> [LockableSqlExpr]
```
parsonsmatt, line 1514: Should this be a `NonEmpty`? Is it reasonable to have a `[]` value for this?

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1983,10 +2083,30 @@ instance Semigroup LimitClause where
 
 instance Monoid LimitClause where
   mempty = Limit mzero mzero
-  mappend = (<>)
 
 -- | A locking clause.
-type LockingClause = Monoid.Last LockingKind
```
parsonsmatt, line 1989: Hmm, changing this type is also probably a breaking change.

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1983,10 +2083,30 @@ instance Semigroup LimitClause where
...
+data LockingClause =
+  GeneralLockingClause LockingKind
+  -- ^ Locking clause not specific to any database implementation
+  | PostgresLockingClause PostgresLockingKind
+  -- ^ Locking clause specific to postgres
```
parsonsmatt, line 2092: Ahh, so that's why we can't hoist the `PostgresLockingKind` to the PG module 🤔 

I think that's probably fine - `MySQL`'s locking clause can be different and in here, too, if we need to create it.

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1405,26 +1426,105 @@ data Insertion a
 -- are no guarantees that they will behave the same.
 --
 -- @since 2.2.7
-data LockingKind
-    = ForUpdate
+data LockingKind where
```
parsonsmatt, line 1429: Can probably more easily write the deriving clause with non-GADT syntax

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1983,10 +2083,30 @@ instance Semigroup LimitClause where
...
+  GeneralLockingClause LockingKind
+  -- ^ Locking clause not specific to any database implementation
+  | PostgresLockingClause PostgresLockingKind
+  -- ^ Locking clause specific to postgres
+  | NoLockingClause
+
+-- | Monoid instance prioritization is:
+-- 1. "Strongest" lock (FOR UPDATE > FOR SHARE) if comparing two postgres
+-- specific locks of differing strength
+-- 2. Rightmost (last) locking clause otherwise
```
parsonsmatt, line 2098: 🤔 

This behavior feels weird. Are the `LockingKind` constructors all more strict than the `PostgresLockingClause`? If I wrote `locking $ forUpdateOfSkipLocked myTable` and composed it with a query that had something more permissive, it would be weird for it to be something more permissive, only because it used a generic constructor instead of a Postgres constructor.

parsonsmatt (commented):
Some bits to fiddle with, I think, if we're going to try and keep this a minor version bump.

Having the `Postgres` specific locking behavior in the `Internal` module is a bit unfortunate. I'd love to get that separated out. Not sure the best way to do that - generally "open world" stuff means "a type class", and stuffing it in a container means "an existential GADT."

src/Database/Esqueleto/Internal/Internal.hs
josephsumabat: This is only used internally to wrap `LockingKind` which is the external one - I could alternatively wrap this one instead?

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -3053,13 +3173,29 @@ makeLimit (conn, _) (Limit ml mo) =
...
+      plain "FOR UPDATE OF " <> makeLockableEntity info lockableEnts
+    makeLockingStrength info (PostgresForShare Nothing) = plain "FOR SHARE"
+    makeLockingStrength info (PostgresForShare (Just (Of lockableEnts))) =
+      plain "FOR SHARE OF " <> makeLockableEntity info lockableEnts
+    makeLockingBehavior :: OnLockedBehavior -> (TLB.Builder, [PersistValue])
+    makeLockingBehavior NoWait = plain "NO WAIT"
+    makeLockingBehavior SkipLocked = plain "SKIP LOCKED"
+    makeLockingBehavior Wait = plain ""
+    plain v = (v,[])
+makeLocking _ NoLockingClause = mempty
```
parsonsmatt, line 3198: So, if we're gonna make this 'extensible', you'd probably need this to be a class member -

```haskell
class ToLockingClause a where
    renderLockingClause :: IdentInfo -> a -> (TLB.Builder, [PersistValue])

data SomeLockingClause where
    SomeLockingClause :: ToLockingClause a => a -> SomeLockingClause
```

Then the stuff in `Internal` can be "generic SQL", and `.Postgres` can define

```haskell
data PostgresqlLockingClause

instance ToLockingClause PostgresqlLockingClause
```

But that's not quite enough. We need a way of determining the "strength", or a way of comparing two locking clauses. Otherwise we can't `<>` two things together.

And comparing two things of potentially disparate types is not really an easy problem to solve!

🤔

src/Database/Esqueleto/Internal/Internal.hs
josephsumabat: Hmm the reason I did it this way is because I didn't want to change the existing behavior of the general locking clause - I could potentially prioritize the more "specific" instance and then within general locking kinds take the rightmost and within postgres take the strongest?

src/Database/Esqueleto/Internal/Internal.hs
josephsumabat: So I tried implementing this but it makes the monoid instance impossible to define I think since I can't pattern match on it :(

I've made it a class member but this function still needs to pattern match to destructure  it.. Not sure if there's a good way of using the existential type and having a monoid instance

src/Database/Esqueleto/Internal/Internal.hs
josephsumabat: Changed to `NonEmpty`!

src/Database/Esqueleto/Internal/Internal.hs
josephsumabat: Ah this was an artifact from the previous implementation, fixed thanks!

src/Database/Esqueleto/Internal/Internal.hs
josephsumabat: Used the derived instance instead!

src/Database/Esqueleto/Internal/Internal.hs
josephsumabat: Added!

src/Database/Esqueleto/Experimental/From/Join.hs
josephsumabat: Added the re-export!

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1425,6 +1449,121 @@ data LockingKind
...
+  --        compareLockStrength (unionLocks a b) b = GT
+  --        compareLockStrength a (unionLocks a b) = EQ
+  --    If compare a b = EQ
+  --        compareLockStrength (unionLocks a b) b = EQ
+  --        compareLockStrength a (unionLocks a b) = EQ
+  --    If compare a b = LT
+  --        compareLockStrength (unionLocks a b) b = EQ
+  --        compareLockStrength a (unionLocks a b) = LT
+  unionLocks :: ComparableLockStrength a => a -> a -> a
+  compareLockStrength :: ComparableLockStrength a => a -> a -> Ordering
```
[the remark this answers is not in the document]
parsonsmatt: If `compareLockStrength` is intended to be an `Ord`like method, we could do a default signature:

```suggestion
  unionLocks :: a -> a -> a
  compareLockStrength :: a -> a -> Ordering
  default compareLockStrength :: Ord a => a -> a -> Ordering
  compareLockStrength = compare
```

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1425,6 +1449,121 @@ data LockingKind
...
+
+instance ComparableLockStrength a => Semigroup (StrongestLock a) where
+  (<>) (StrongestLock Nothing) pright = pright
+  (<>) pleft (StrongestLock Nothing) = pleft
+  (<>) (StrongestLock (Just pright)) (StrongestLock (Just pleft)) =
+    case compareLockStrength pleft pright of
+      GT -> StrongestLock (Just pright)
+      EQ -> StrongestLock (Just $ unionLocks pright pleft)
+      LT -> StrongestLock (Just pleft)
```
parsonsmatt, line 1478: 🤔 

```suggestion
  (<>) (StrongestLock (Just pleft)) (StrongestLock (Just pright)) =
    case compareLockStrength pleft pright of
      GT -> StrongestLock (Just pright)
      EQ -> StrongestLock (Just $ unionLocks pright pleft)
      LT -> StrongestLock (Just pleft)
```

Which I think also swaps the other ones.

```
> compare 1 2
LT
```

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1425,6 +1449,121 @@ data LockingKind
...
+  compareLockStrength a b = compare (postgresRowLevelLockStrength a) (postgresRowLevelLockStrength b)
+
+-- Arranged in order of lock strength
+data PostgresRowLevelLockStrength =
+    PostgresForUpdate
+    | PostgresForShare
+  deriving (Ord, Eq)
+
+data LockingOfClause where
+  Of :: LockableEntity a => a -> LockingOfClause
```
parsonsmatt, line 1513:
```suggestion
  LockingOfClause :: LockableEntity a => a -> LockingOfClause
```

`Of` is a pretty small word and it would be difficult to track what it's from - sharing the name will make it easier

src/Database/Esqueleto/Internal/Internal.hs
parsonsmatt: Indeed! It's fine 😄

parsonsmatt (approved):
alright, I'm happy with this API! Thanks for working with me to polish it up nice 😄

src/Database/Esqueleto/Internal/Internal.hs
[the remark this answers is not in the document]
josephsumabat: I don't think we want the `Ord` constraint though since we can compare the locking strength of two locks where it doesn't make sense to have an `Ord` instance -> that is it only compare the `postgresRowLevelLockStrength` of `PostgresLockingKind` and not any of the other behavior.

josephsumabat:
@parsonsmatt I've changed the monoid instance to concatenate postgres locking clauses (which results in a valid query) and prioritize general locking clauses over postgres ones.

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1406,7 +1430,7 @@ data Insertion a
 --
 -- @since 2.2.7
 data LockingKind
-    = ForUpdate
+   = ForUpdate
```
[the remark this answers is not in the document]
josephsumabat: fixed!

src/Database/Esqueleto/Internal/Internal.hs
josephsumabat: removed this since we can just concatenate the locking clauses for postgres

josephsumabat:
Past few commits were just updating documentation and version number (since another CR was merged!)

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1983,10 +2077,32 @@ instance Semigroup LimitClause where
...
+  | NoLockingClause
+
+instance Semigroup LockingClause where
+  -- Postgres allows us to have multiple locking clauses
+  (<>) (PostgresLockingClauses pleft) (PostgresLockingClauses pright) = PostgresLockingClauses (pleft <> pright)
+  -- For backwards compatibility reasons we use the monoid instance of last for
+  -- general locking cluase
+  (<>) (GeneralLockingClause gleft) (GeneralLockingClause gright)  =
+    Maybe.maybe mempty GeneralLockingClause
+      (Monoid.getLast $ Monoid.Last (Just gleft) <> Monoid.Last (Just gright))
```
parsonsmatt, line 2096: I think that `getLast (Last _ <> Last b)` will always be `Last b`, so we can probably simplify this 

```suggestion
    GeneralLockingClause gright
```

test/PostgreSQL/Test.hs
```diff
@@ -1227,6 +1232,197 @@ testCommonTableExpressions = do
             pure res
         asserting $ vals `shouldBe` fmap Value [2..11]
 
+testPostgresqlLocking :: SpecDb
+testPostgresqlLocking = do
+    describe "Monoid instance" $ do
+      let toText conn q =
```
parsonsmatt, line 1238: please use 4 space indent thoughout this project

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1983,10 +2077,32 @@ instance Semigroup LimitClause where
...
+
+  (<>) gleft@(GeneralLockingClause _) (PostgresLockingClauses _)  = gleft
+  (<>) (PostgresLockingClauses _) gright@(GeneralLockingClause _) = gright
```
parsonsmatt, line 2099: I feel pretty unsatisfied here. I can imagine that this would be *very* confusing to debug if it comes up in a production context. Imagine looking at a `SqlQuery` that has locking behavior you want, and composing that with another `SqlQuery`. Later on, someone adds a `GeneralLockingClause` to the other query, and suddenly you lose the locking properties in the query you care about. The `GeneralLockingClause` destroys the `PostgresLockingClause` information, and the only information you'd get is that the query silently doesn't do what you want.

🤔 

It's especially challenging because the `GeneralLockingClause` just takes the `Last` behavior - so if you had `foo >> bar`, where `foo` does a `GeneralLockingClause`, then `bar`'s own locking clause would *overwrite* `foo` if it is a `GeneralLockingClause`. But if `bar` is a `PostgresLockingClause`, then it does the opposite thing - `foo` wins and `bar` loses.

Can we always favor the last locking clause - ie,

```haskell
    PostgresLockingClause l <> PostgresLockingClause r = 
        PostgresLockingClause (l <> r)
    l <> NoLockingClause =
        l
    l <> r = 
        r
```

I'm not sure that satisfies the monoid/semigroup laws.

I wonder if it is reasonable to collect a `GeneralLockingClause` in with the `PostgresLockingClause`? I wish I were more familiar with the interaction of these features.

parsonsmatt (commented):
I'm concerned about the way the API drops locking clauses here.

(Fwiw, the prior `Last` implementation *also* feels wrong, but I'm not sure the right way to reconcile that).

I want to do a bit more research on how these work and are accepted syntactically, both by Postgres and other backends.

src/Database/Esqueleto/Internal/Internal.hs
josephsumabat: I've shifted to taking the `Last` when choosing between general and postgres locking clauses (aggregating them together doesnt work well since some of the "general" ones are rdbms specific and consequently unsafe- I've renamed it `LegacyLockingClause` to better reflect that and I would probably argue in a future release we only allow one type of locking clause)

test/PostgreSQL/Test.hs
josephsumabat: Thanks, fixed!

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1983,10 +2078,24 @@ instance Semigroup LimitClause where
...
-  mappend = (<>)
 
 -- | A locking clause.
-type LockingClause = Monoid.Last LockingKind
+data LockingClause =
+    LegacyLockingClause LockingKind
+    -- ^ Locking clause not specific to any database implementation
+    | PostgresLockingClauses [PostgresLockingKind]
+    -- ^ Locking clause specific to postgres
+    | NoLockingClause
```
parsonsmatt, line 2088: Can we put `@since` annotations on these?

parsonsmatt (approved):
OK, I'm happy with this for now - Sorry for taking so long to get a review on it.

Going to go ahead and merge, and will edit the version to 3.5.9.0

src/Database/Esqueleto/Internal/Internal.hs
josephsumabat: I believe `LockingClause` is internal only - do you still want me to add?

PR #342, opened 2022-11-08
`TypeError` instance for `Functor SqlExpr`

parsonsmatt:

- [~] Ran `stylish-haskell` and otherwise adhered to the [style guide](https://github.com/bitemyapp/esqueleto/blob/master/style-guide.yaml).

After submitting your PR:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,10 @@
+3.5.8.2
+=======
+- @parsonsmatt
+    - []()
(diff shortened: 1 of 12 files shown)


ivb-supercede:
Want to say that I really support this change. I think it would have been very helpful when I was starting out using the library, since I was sometimes confused about in which layer code was being executed.

PR #350, opened 2023-02-16
Add GetFirstTable

9999years:

After submitting your PR:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,9 @@
-3.5.9.2
-=======
+3.5.9.0
+========
+- @9999years
+    - #350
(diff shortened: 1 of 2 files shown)


belevy:
Perhaps we should allow manual aliasing and make something like `NamedTable "p1" Person` and then you could use the symbol to differentiate

9999years:
@belevy That sounds very neat — how would users assign labels to each table?

parsonsmatt (approved):
Looks great! Let's bump version + `@since` annotations and get this out there

parsonsmatt (approved):

Issue #345, opened 2022-11-15
Support [NOT] MATERIALIZED keyword for WITH queries/CTEs


joelmccracken:
As mentioned in the esquelto docs (https://hackage.haskell.org/package/esqueleto-3.5.8.1/docs/Database-Esqueleto-Experimental-From-CommonTableExpression.html), postgres (>12) supports keywords which specifies that CTEs should be fully calculated/materialized. For us, this is a must for performance reasons.

It would be great if there was some way to do this in esquelto, so that we could use it everywhere instead of having to use `sqlQQ` in cases where we need to specify materialized views.

belevy:
Is there a reason to use a cte if you don't need the materialization? Can you not just use a Haskell variable for the subquery?

belevy:
Sorry I misunderstood the question. Yes postgres module could have a with that specifies materialized

joelmccracken:
Thinking about working on this soon, any advice?

joelmccracken:
So, my initial thoughts was to add another element to `CommonTableExpressionKind`, like so:

```
data CommonTableExpressionKind
    = RecursiveCommonTableExpression
    | NormalCommonTableExpression
    | MaterializedCommonTableExpression
    deriving Eq
```

However, thinking this through, I realized that this has a problem: it omits the case where there are recursive AND materialized CTEs.

I guess for now I'm going to add another paraemter somewhere and make a `withMaterialized` and `withRecursiveMaterialized` functions, something like that.

PR #354, opened 2023-03-07
Postgres Materialized CTE functions

joelmccracken:
Adds `withMaterialized` function to support this non-standard postgres behavior. 

After submitting your PR:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,9 @@
+3.5.12.0
+========
+- @JoelMcCracken
+    - [#354](https://github.com/bitemyapp/esqueleto/pull/354)
+      - Add `withMaterialized`, `withNotMaterialized` to the PostgreSQL module
+
 3.5.11.2
 ========
 - @arguri
--- a/esqueleto.cabal
+++ b/esqueleto.cabal
@@ -2,7 +2,7 @@ cabal-version: 1.12
 
 name:           esqueleto
 
-version:        3.5.11.2
+version:        3.5.12.0
 synopsis:       Type-safe EDSL for SQL queries on persistent backends.
 description:    @esqueleto@ is a bare bones, type-safe EDSL for SQL queries that works with unmodified @persistent@ SQL backends.  Its language closely resembles SQL, so you don't have to learn new concepts, just new syntax, and it's fairly easy to predict the generated SQL and optimize it for your backend. Most kinds of errors committed when writing SQL are caught as compile-time errors---although it is possible to write type-checked @esqueleto@ queries that fail at runtime.
                 .
--- a/src/Database/Esqueleto/Experimental/From/CommonTableExpression.hs
+++ b/src/Database/Esqueleto/Experimental/From/CommonTableExpression.hs
@@ -38,7 +38,8 @@ import Database.Esqueleto.Internal.Internal hiding (From(..), from, on)
 -- PostgreSQL 12, non-recursive and side-effect-free queries may be inlined and
 -- optimized accordingly if not declared @MATERIALIZED@ to get the previous
 -- behaviour. See [the PostgreSQL CTE documentation](https://www.postgresql.org/docs/current/queries-with.html#id-1.5.6.12.7),
--- section Materialization, for more information.
+-- section Materialization, for more information. To use a @MATERIALIZED@ query
+-- in Esquelto, see functions 'withMaterialized' and 'withRecursiveMaterialized'.
 --
 -- /Since: 3.4.0.0/
 with :: ( ToAlias a
@@ -50,7 +51,7 @@ with query = do
     aliasedValue <- toAlias ret
     let aliasedQuery = Q $ W.WriterT $ pure (aliasedValue, sideData)
     ident <- newIdentFor (DBName "cte")
-    let clause = CommonTableExpressionClause NormalCommonTableExpression ident (\info -> toRawSql SELECT info aliasedQuery)
+    let clause = CommonTableExpressionClause NormalCommonTableExpression (\_ _ -> "")  ident (\info -> toRawSql SELECT info aliasedQuery)
     Q $ W.tell mempty{sdCteClause = [clause]}
     ref <- toAliasReference ident aliasedValue
     pure $ From $ pure (ref, (\_ info -> (useIdent info ident, mempty)))
@@ -103,7 +104,8 @@ withRecursive baseCase unionKind recursiveCase = do
     ref <- toAliasReference ident aliasedValue
     let refFrom = From (pure (ref, (\_ info -> (useIdent info ident, mempty))))
     let recursiveQuery = recursiveCase refFrom
-    let clause = CommonTableExpressionClause RecursiveCommonTableExpression ident
+    let noModifier _ _ = ""
+    let clause = CommonTableExpressionClause RecursiveCommonTableExpression noModifier ident
                  (\info -> (toRawSql SELECT info aliasedQuery)
                         <> ("\n" <> (unUnionKind unionKind)  <> "\n", mempty)
                         <> (toRawSql SELECT info recursiveQuery)
--- a/src/Database/Esqueleto/PostgreSQL.hs
+++ b/src/Database/Esqueleto/PostgreSQL.hs
@@ -34,6 +34,8 @@ module Database.Esqueleto.PostgreSQL
     , forShareOf
     , filterWhere
     , values
+    , withMaterialized
+    , withNotMaterialized
     -- * Internal
     , unsafeSqlAggregateFunction
     ) where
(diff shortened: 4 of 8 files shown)


joelmccracken:
fixes #345

belevy:
Is there a way that we can avoid more postgres specific stuff in the internal module? Since `NOT MATERIALIZED` comes after the `AS` could we just prepend it to the query? Perhaps we just need a more generic `type CTEModifier = IdentInfo -> TLB.Builder`

belevy:
Yah so the whole point is to keep that modifier text open for extension. Its obviously less safe from an implementation perspective but who is to say that MySQL won't come out with their own custom modifiers. You have a couple of options here that are all pretty distateful each in their own way. 

1. Strict seperation of the SqlQuery type by backend. Each backend has a completely unique SqlQuery type, this is then the subject of typeclasses to support code reuse. This is trading strict correctness checking in the types for potential duplication in the implementation as much of the query language is standardized with a few changes.

2. Parameterize a SqlQuery by its constituent parts (i.e. data SqlQuery cteClause fromClause whereClause ... =). This has the nice effect of letting you ignore the differing parts when they don't matter. It however still has some major duplication issues that show up the minute you need to make a custom clause type and for total safety you will end up needing to create custom clauses for each part.

3. Create a single unified type that has all of the cases for all of the backends baked in, this is what esqueleto generally has done in the past. You have a single sum type that needs to handle all of the different possible backends. This is super simple but has a nasty downside in that it drags everything into the internal module, since the type is not open for extension from a different module we end up needing to modify the mega sum type, this also means that you don't really have safety to prevent mixing statements that are specific to different backends and also means you have to rely on all of the backends always (it is currently impossible to break esqueleto-core, esqueleto-postgres, esqueleto-mysql, and esqueleto-sqlite out into independent packages). 

4. The approach that I have been tending towards for use in libraries lately which is something akin to a monomorphic tagless final. In this approach there is a single SqlQuery type that is the final representation of a query (or near enough) and has extension points where the backends differ. This way we can share almost all of the code and only have to change exactly what differs, this still requires the internal module to know about how backends might wish to customize the core type but it frees us of the compile time dependency. This approach is very much like OOP where you can envision all of the functions producing `SqlQuery` as classes that implement a `SqlQuery` interface ala the interpreter pattern. This has a nasty downside of removing compile time help from the implementer but I have felt that it ended up making far simpler code when I switched the Experimental module to this style. 

So to answer your question, yes leaving the modifier as a free for all is kind of icky but the end user won't see it only the implementer.

joelmccracken:
@belevy thank you for explaining. I think I got it, and have addressed your comments.

Let me know if there is anything else to address!

joelmccracken:
I ran stylish-haskell on the files that were modified in this PR so I could mark off that check list item, just in case it caused anyone pause.  For reference, however, it failed for me on `Internal.hs`.

```
for f in $(git diff --name-only origin/master | grep '.hs'); do stack exec stylish-haskell -- -i $f || continue ; done
src/Database/Esqueleto/Internal/Internal.hs: RealSrcSpan SrcSpanPoint "src/Database/Esqueleto/Internal/Internal.hs" 427 1: lexical error in string/character literal at character 's'
```

src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -477,3 +486,97 @@ forUpdateOf lockableEntities onLockedBehavior =
...
+-- where_ $ cteResult ...
+-- pure cteResult
+-- @
+--
+--
+-- For more information on materialized CTEs, see the PostgreSQL manual documentation on
+-- [Common Table Expression Materialization](https://www.postgresql.org/docs/14/queries-with.html#id-1.5.6.12.7).
+--
+-- /Since: 3.5.10.0/
+withMaterialized :: ( ToAlias a
```
belevy, line 512: We probably want to also support non materialized so perhaps the materialization modifier can be passed in as an argument and we can provide helpers `materialized` and `notMaterialized` to safely select from. In fact, the more general versions could be in the core module and then whether we export with as `withInternal noModifier` or `withInternal` is whether you use with from the core module or with from this module.

belevy (changes requested):
Thank you for moving the postgresql specifics out of Internal. I feel bad requesting more changes, especially after how long I took reviewing. I think we are getting close, perhaps @parsonsmatt has opinions but I feel like with the exception of the few notes I made it seems pretty good.

src/Database/Esqueleto/PostgreSQL.hs
joelmccracken: So here was what I was thinking with this PR:

- `with`, in Database/Esqueleto/Experimental/From/CommonTableExpression.hs, is
  usable for any of the sql backends; it doesn't do anything with materialized.
  - If a postgres user wants to just use regular `with` CTE and have postgres
    figure out if it wants to materialize it or not, they can just use `with`
    from the normal module.
- the `MATERIALIZED` modifier is only supported in postgres, so the ability to
  specify this behavior is in the pg specific module.
  - If a pg user wishes to force pg to materialize the cte, they must import the
    pg module and use `withMaterialized`
- But, besides `MATERIALIZED`, there is also a `NOT MATERIALIZED` possible
  modifier, which I have *not* added support for in this version . I can add
  it if desired, however, but I'll discuss why I didn't do it below.

Given that, when you say:

> We probably want to also support non materialized

Are you referring to when *no* materialization modifier is specified, or when
the `NOT MATERIALIZED` modifier is intended to be used? Either way is fine; it
makes sense to include something that would do `NOT MATERIALIZED`, for
completeness.

--- 

On the topic of supporting the `NOT MATERIALIZED` modifier, here is what my
thought process was:

I saw the following function pattern when investigating this feature:
- `with` 
- `withRecursive` 

To me, this implied I should add these functions:
- `withMaterialized`
- `withMaterializedRecursive`

But, for `NOT MATERIALIZED`, following the pattern would result in even more
functions (and this just seemed like too many):
- `withNotMaterialized`
- `withNotMaterializedRecursive`

So, I didn't add them; totally fine with adding them, I just want to make sure
we're talking about the same thing. 

--- 

Also, I just re-examined the docs for pg 12-15, and I notice that recursive CTEs are
_always_ "materialized", so `withMaterializedRecursive` and
`withNotMaterializedRecursive` are superfluous. FYI, according to my tests
with pg 15, all of these combinations actually _function_, even if they all
actually behave in the same way. 

---

So, given the above, I *think* the API should be one of these two choices:

- first scenario, I delete `withMaterializedRecursive` and add `withNotMaterialized`
- second scenario, add modifiers as arguments
  - pg module would have
    - `with` function having *two* arguments
      - first argument is `Maybe Materilalization`
        - `data Materilization = Materialized | NotMaterialized`
        - So three possible settings,
          - `None` for providing _no_ materialized modifier
          - `Just Materialized` for providing `MATERIALIZED` modifier
          - `Just NotMaterialized` for providing `NOT MATERIALIZED` modifier
      - the second argument is `query`, as it is today
    - `withRecursive` is deleted 
  - the regular `with` function should have the same API as today in Experimental.From.CommonTableExpression.

Thoughts?

joelmccracken: @belevy I went ahead and made this change anyway

src/Database/Esqueleto/Experimental/From/CommonTableExpression.hs
```diff
@@ -38,7 +38,8 @@ import Database.Esqueleto.Internal.Internal hiding (From(..), from, on)
 -- PostgreSQL 12, non-recursive and side-effect-free queries may be inlined and
 -- optimized accordingly if not declared @MATERIALIZED@ to get the previous
 -- behaviour. See [the PostgreSQL CTE documentation](https://www.postgresql.org/docs/current/queries-with.html#id-1.5.6.12.7),
--- section Materialization, for more information.
+-- section Materialization, for more information. To use a @MATERIALIZED@ query
+-- in Esquelto, see functions 'withMaterialized' and 'withRecursiveMaterialized'.
```
parsonsmatt, line 42:
```suggestion
-- in Esqueleto, see functions 'withMaterialized' and 'withRecursiveMaterialized' defined in 
-- "Database.Esqueleto.PostgreSQL".
```

src/Database/Esqueleto/PostgreSQL.hs
parsonsmatt: I'm pretty happy with the API that is presented in this PR.

esqueleto.cabal
```diff
@@ -2,7 +2,7 @@ cabal-version: 1.12
 
 name:           esqueleto
 
-version:        3.5.11.2
+version:        3.5.12.0
```
parsonsmatt, line 5:
```suggestion
version:        3.5.14.0
```

changelog.md
```diff
@@ -1,3 +1,9 @@
+3.5.12.0
```
```suggestion
3.5.14.0 (unreleased)
```

parsonsmatt (approved):
This looks great to me - thanks! I am planning on releasing 3.5.14.0 today and should have this incorporated.

parsonsmatt:
Closing in favor of https://github.com/bitemyapp/esqueleto/pull/417

Issue #306, opened 2022-03-23
Improved Postgres Update


parsonsmatt:
This is going to be a mega issue that documents improving `update` for `Postgresql` specifically. 

Related issues: 

- [ ] #251 `update` doesn't work with experimental syntax
- [ ] #204 Add a `WHERE` to `insertSelectOnConflict`
- [ ] #44 Support `update ... returning`

# Postgres Docs

[These are the Postgres docs for `update`](https://www.postgresql.org/docs/current/sql-update.html). Copying the syntax, we have:

```
[ WITH [ RECURSIVE ] with_query [, ...] ]
UPDATE [ ONLY ] table_name [ * ] [ [ AS ] alias ]
    SET { column_name = { expression | DEFAULT } |
          ( column_name [, ...] ) = [ ROW ] ( { expression | DEFAULT } [, ...] ) |
          ( column_name [, ...] ) = ( sub-SELECT )
        } [, ...]
    [ FROM from_item [, ...] ]
    [ WHERE condition | WHERE CURRENT OF cursor_name ]
    [ RETURNING * | output_expression [ [ AS ] output_name ] [, ...] ]
```

Right now, our `update` function in `esqueleto` only supports updating a single table (no `FROM` clause allowed), and we don't have the ability to return anything from it aside from the count of rows modified.

# A type and API

- `UPDATE ... RETURNING` returns a list of rows that were actually updated, so it should have a return type like `select`.
- The `SqlQuery` already allows us to do CTEs, but we'll have to be careful to structure them outside the `UPDATE` syntax.
- The `FROM` stuff is a bit tricky. Per docs,
    > When a FROM clause is present, what essentially happens is that the target table is joined to the tables mentioned in the from_item list, and each output row of the join represents an update operation for the target table. When using FROM you should ensure that the join produces at most one output row for each row to be modified. In other words, a target row shouldn't join to more than one row from the other table(s). If it does, then only one of the join rows will be used to update the target row, but which one will be used is not readily predictable.
- More `FROM` trickiness - we can't update non-target tables. So the type introduced by the lambda should be slightly different from the type introduced by `from`.

The current signature is:

```haskell
update :: (SqlExpr (Entity val) -> SqlQuery ()) -> SqlWriteT m ()
```

To account for the above, we want to do something like:

```haskell
update
    :: (SqlSelect a r, PersistEntity val, _)
    => (UpdateTarget val -> SqlQuery a)
    -> SqlWriteT m [r]
```

[...]

Issue #44, opened 2017-07-31
Support UPDATE ... RETURNING


jfischoff:
At least with [postgres](https://www.postgresql.org/docs/current/static/sql-update.html), one can add a `RETURNING` clause to an `UPDATE` query. The type in `esqueleto` does not support returning data from [`update`](https://hackage.haskell.org/package/esqueleto-2.5.3/docs/Database-Esqueleto.html#v:update)

parsonsmatt:
I think the way we can do this is:

```haskell
updateReturning
  :: (SqlExpr (Entity val) -> SqlQuery ret)
  -> SqlWriteT m ret
```

defined in `Database.Esqueleto.PostgreSQL`.

ulidtko:
If anyone has alternative ways to reach @parsonsmatt or other maintainers — please ping them. 

This has been implemented 6 releases ago — PR #362 — but it didn't receive any review so far.

PR #362, opened 2023-05-11
Support UPDATE/DELETE .. RETURNING

ulidtko:
Howdy!

This implements #44 :tada: The aggregate issue #306 is also related.

`beginner friendly` indeed :ok_hand:  was a bit scary at first, but I found the way! Pretty pleased with the result.

Please review & merge :pray: 

--------------
PR template checklist:

- [ ] Ran `stylish-haskell` and otherwise adhered to the [style guide](https://github.com/bitemyapp/esqueleto/blob/master/style-guide.yaml)
    :warning: this fails with syntax error:
    > src/Database/Esqueleto/Internal/Internal.hs: RealSrcSpan SrcSpanPoint "src/Database/Esqueleto/Internal/Internal.hs" 427 1: lexical error in string/character literal at character 's'

    BTW that weird `sub_select` deprecation notice is also breaking VS Code syntax highlighting, badly. Vim handles it, but also emits a parse error:
  image

After submitting your PR:


ulidtko:
Gee, `INSERT` also supports `RETURNING`

ulidtko:
Changelog rebased after v3.5.10.0 release. @parsonsmatt @ivanbakel please review

ulidtko:
Nicely wrapping `INSERT .. RETURNING` as well in this DSL is above my head at this point. :confused: 

I'd tried several times; the best I got is a failing test that generates the expected query:

```
[Debug#SQL] INSERT INTO "Person"("name","age","weight","favNum") VALUES(?,?,?,?) RETURNING "id"; [PersistText "John",PersistInt64 36,PersistNull,PersistInt64 1]
```

but then result readout fails with `UnsupportedSqlInsertIntoType`, apparently here:

https://github.com/bitemyapp/esqueleto/blob/e50fed1ddb2f3a6d5f148c5f9148a58517c9d4a8/src/Database/Esqueleto/Internal/Internal.hs#L3302

But then to use the other `SqlSelect` instance with `Insertion a`, I'd need to fix its `sqlSelectProcessRow`, and then somehow combine `rawSelectSource` and `rawEsqueletto` to both be passing input data "into" the query, and get the conduit for reading "out" of it.

--------------------

Factoring in the complete radio silence from maintainers — I won't invest much further effort.

ulidtko:
Ping @parsonsmatt @9999years @arguri @halogenandtoast @isomorpheme @ttuegel

image

parsonsmatt:
Hey, I'm sorry it's taken me a long time to look at this. I'll review it today.

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1835,14 +1833,15 @@ data SideData = SideData
     , sdLimitClause    :: !LimitClause
     , sdLockingClause  :: !LockingClause
     , sdCteClause      :: ![CommonTableExpressionClause]
+    , sdReturningClause :: !ReturningClause
```
parsonsmatt, line 1836: I wonder if `Maybe ReturningClause` is a better type here. Given a satisfactory `Semigroup ReturningClause`, we get the desired `Monoid` for free from `Maybe`.

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -1879,6 +1878,10 @@ data CommonTableExpressionKind
 data CommonTableExpressionClause =
     CommonTableExpressionClause CommonTableExpressionKind Ident (IdentInfo -> (TLB.Builder, [PersistValue]))
 
+data ReturningClause
+    = ReturningNothing  -- ^ The default, absent clause.
+    | ReturningStar     -- ^ @RETURNING@ is present.
```
parsonsmatt, line 1883: If we use `Maybe ReturningClause` as the `SideData` type, then this  has one fewer case:

```suggestion
data ReturningClause
    = ReturningStar     -- ^ @RETURNING@ is present.
```

But `ReturningStar` itself isn't carring any useful information - we could further fold that `Maybe ReturningClause` into just a `Bool`.

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -3268,6 +3318,12 @@ makeLocking info (PostgresLockingClauses clauses) =
             plain v = (v,[])
 makeLocking _ NoLockingClause = mempty
 
+makeReturning :: SqlSelect a r
+              => IdentInfo -> ReturningClause -> a -> (TLB.Builder, [PersistValue])
+makeReturning _    ReturningNothing _ = mempty
+makeReturning info ReturningStar  ret = ("RETURNING ", []) <> sqlSelectCols info ret
```
parsonsmatt, line 3324: This is getting the `SqlSelect a r` instance and using that to provide `sqlSelectCols info ret` here.

A typical `update` in `esqueleto` has type `update :: (SqlExpr (Entity val) -> SqlQuery ()) -> SqlPersistT m ()`. This would give us an `a` of `()`, which would use this instance:

```haskell
-- | Not useful for 'select', but used for 'update' and 'delete'.
instance SqlSelect () () where
    sqlSelectCols _ _ = ("1", [])
    sqlSelectColCount _ = 1
    sqlSelectProcessRow _ = Right ()
```

`RETURNING 1`.

*If* this clause `ReturningStar` is enabled in a *non-update/insert/delete* - ie,

```haskell
oops = select do
  t <- from $ table @Foo
  tellReturning ReturningStar
  pure t
```

Then we'll generate a query like:

```sql
SELECT foo.a, foo.b, foo.c
FROM foo
RETURNING foo.a, foo.b, foo.c
```

Which is definitely an error.

So the possibility of error here makes me feel like exploring a different approach.

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -3954,7 +4010,7 @@ renderExpr :: SqlBackend -> SqlExpr (Value Bool) -> T.Text
 renderExpr sqlBackend e = case e of
     ERaw _ mkBuilderValues ->
          let (builder, _) = mkBuilderValues Never (sqlBackend, initialIdentState)
-         in (builderToText builder)
+         in builderToText builder
```
parsonsmatt, line 4013: Thanks for the drive-by improvements!

src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -477,3 +480,49 @@ forUpdateOf lockableEntities onLockedBehavior =
...
+-- @
+--
+-- @since 3.5.11.3
+updateReturning :: (MonadIO m, From from, InferReturning ex ret, SqlBackendCanWrite backend)
+                => (from -> SqlQuery ex)
+                -> R.ReaderT backend m [ret]
+updateReturning block = do
+  conn <- R.ask
+  conduit <- rawSelectSource UPDATE (tellReturning ReturningStar >> from block)
+  liftIO . withAcquire conduit $ flip R.runReaderT conn . runSource
```
parsonsmatt, line 506: So this is the *use* of the new functionality - neat.

This is where we say "Do the `RETURNING` thing":

```haskell
rawSelectSource UPDATE (tellReturning ReturningStar >> from block)
```

So we modify the `SideData` for the query, then it gets rendered and turned into a `conduit`, which we run.

I think we can *avoid* modifying `SideData` and still mostly keep this interface.

Consider [`renderQueryToText`](https://github.com/bitemyapp/esqueleto/blob/30a5e80736391e2aa45094f681d4bd329aa16707/src/Database/Esqueleto/Internal/Internal.hs?plain=1#L3005-L3024). This takes a `SqlQuery a` and renders it out.

If the `SqlQuery a` has a `ReturningStar` in the `SideData`, then we graft on the appropriate text to the query. However, we can just *do that* directly without modifying the internals and introducing potential unsafety.

If we look at the code for [`rawSelectSource`](https://github.com/bitemyapp/esqueleto/blob/30a5e80736391e2aa45094f681d4bd329aa16707/src/Database/Esqueleto/Internal/Internal.hs?plain=1#L2717-L2745), we can see that it's just doing `toRawSql` on the query and then `rawQueryRes`. The rest is handling the `Conduit` (which imo isn't super useful since the query must be loaded into memory anyway).

So, if we inline this and use `rawQuery` instead (which just does the list directly), we can do:

```haskell
updateReturning block = do
    conn <- R.ask
    let 
        query' = toRawSql UPDATE (conn, initialIdentState) block
        ((ret, _sd), _finalIdent) = flip S.runState -- ... (from toRawSql) 
        returningClause =  ("RETURNING ", []) <> sqlSelectCols info ret
        (queryWithReturning, vals) = 
             query' <> ("\n", []) <> returningClause
    rawQuery queryWithReturning vals
```

Now, it *is* unsatisfying that we have to run the query building twice with this - once in `toRawSql` and once to get the `ret` to make the `returningClause` with.

If `toRawSql` is refactored into a variant that *also* returns the `ret`, then I think we're able to avoid that - then this can be:

```haskell
    conn <- R.ask
    let 
        (ret, query') = toRawSqlWith UPDATE (conn, initialIdentState) block
        returningClause =  ("\nRETURNING ", []) <> sqlSelectCols info ret
        (queryWithReturning, vals) = 
             query' <> returningClause
    rawQuery queryWithReturning vals
```

Then I think there's probably virtue in writing `makeReturning :: (InferReturning a r) => a -> (TLB.Builder, [PersistValue])` which can then be reused in `deleteReturning`.

test/PostgreSQL/Test.hs
```diff
@@ -1056,6 +1056,38 @@ testUpsert =
...
+        where_ (isNothing $ p ^. PersonWeight)
+        return (val (1 :: Int, 2 :: Int), p ^. PersonName, isNothing (p ^. PersonAge))
+      asserting $ ret1 `shouldBe` [ (Value (1, 2), Value "John", Value False)
+                                  , (Value (1, 2), Value "Mike", Value False)
+                                  , (Value (1, 2), Value "Mitch", Value True) ]
+      ret2 <- EP.deleteReturning $ \p -> do
+        -- empty WHERE -- delete everything remaining... but:
+        return (p, p ^. PersonName)
+      asserting $ ret2 `shouldBe` [ (Entity p2k p2, Value "Rachel")
+                                  , (Entity p4k p4, Value "Livia") ]
```
parsonsmatt, line 1089: Very nice tests, thank you!

changelog.md
```diff
@@ -1,3 +1,9 @@
+3.5.11.3 (unreleased)
+========
+- @ulidtko
+    - [#362](https://github.com/bitemyapp/esqueleto/pull/362)
+      - Add `updateReturning`, `deleteReturning` as Postgres extensions.
```
parsonsmatt, line 5: This implementation adds a field to the `SideData` type, which unfortunately does make this a breaking change and would need to go out as `3.6.0.0`.

However, I've suggested an alternative implementation that should be strictly additive, allowing it to be `3.5.12.0`.

src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -2400,6 +2412,42 @@ existsHelper = sub SELECT . (>> return true)
...
+         , InferReturning c rc
+         , InferReturning d rd
+         ) => InferReturning (a, b, c, d) (ra, rb, rc, rd)
+instance ( InferReturning a ra
+         , InferReturning b rb
+         , InferReturning c rc
+         , InferReturning d rd
+         , InferReturning e re
+         ) => InferReturning (a, b, c, d, e) (ra, rb, rc, rd, re)
+instance ( InferReturning a ra
```
parsonsmatt, line 2442: I'd prefer to see a newline between the instance declarations
```suggestion
class SqlSelect a r => InferReturning a r | r -> a, a -> r

instance PersistEntity ent => InferReturning (SqlExpr (Entity ent)) (Entity ent)

instance PersistEntity ent => InferReturning (SqlExpr (Maybe (Entity ent))) (Maybe (Entity ent))

instance PersistField a => InferReturning (SqlExpr (Value a)) (Value a)

instance ( InferReturning a ra, InferReturning b rb) => InferReturning (a, b) (ra, rb)

instance ( InferReturning a ra
         , InferReturning b rb
         , InferReturning c rc
         ) => InferReturning (a, b, c) (ra, rb, rc)
         
instance ( InferReturning a ra
         , InferReturning b rb
         , InferReturning c rc
         , InferReturning d rd
         ) => InferReturning (a, b, c, d) (ra, rb, rc, rd)
         
instance ( InferReturning a ra
         , InferReturning b rb
         , InferReturning c rc
         , InferReturning d rd
         , InferReturning e re
         ) => InferReturning (a, b, c, d, e) (ra, rb, rc, rd, re)
         
instance ( InferReturning a ra
```

src/Database/Esqueleto/PostgreSQL.hs
```diff
@@ -477,3 +480,49 @@ forUpdateOf lockableEntities onLockedBehavior =
...
+-- @
+--
+-- @since 3.5.11.3
+deleteReturning :: (MonadIO m, From from, InferReturning ex ret, SqlBackendCanWrite backend)
+                => (from -> SqlQuery ex)
+                -> R.ReaderT backend m [ret]
+deleteReturning block = do
+  conn <- R.ask
+  conduit <- rawSelectSource DELETE (tellReturning ReturningStar >> from block)
+  liftIO . withAcquire conduit $ flip R.runReaderT conn . runSource
```
parsonsmatt, line 528: Style nit: please use 4 space indentation.

```suggestion
    conn <- R.ask
    conduit <- rawSelectSource DELETE (tellReturning ReturningStar >> from block)
    liftIO . withAcquire conduit $ flip R.runReaderT conn . runSource
```

test/PostgreSQL/Test.hs
```diff
@@ -1056,6 +1056,38 @@ testUpsert =
...
+testUpdateDeleteReturning :: SpecDb
+testUpdateDeleteReturning =
+  describe "UPDATE .. RETURNING .." $ do
+    itDb "Whole entities, expressions and tuples get returned from UPDATE" $ do
+      [_p1k, _p2k, _p3k, p4k, _p5k] <- mapM insert [p1, p2, p3, p4, p5]
+      ret1 <- EP.updateReturning $ \p -> do
+        set p [ PersonFavNum =. val 42 ]
+        where_ (p ^. PersonFavNum ==. val 4)
+        return p
+      asserting $ ret1 `shouldBe` [Entity p4k p4{ personFavNum = 42 }]
```
parsonsmatt, line 1068: 4 space indentation here too please 🙏🏻

parsonsmatt (changes requested):
Thanks for the contribution! I apologize again for taking so  long to review it.

I've suggested an alternative implementation which should be safer and possibly even slightly simpler. The backbone comes from refactoring `toRawSql` to expose the `ret` from the `SqlQuery`, which allows you to append the `RETURNING` clause directly and avoid modifying `SideData`:

```haskell
-- | (Internal) Pretty prints a 'SqlQuery' into a SQL query.
--
-- Note: if you're curious about the SQL query being generated by
-- @esqueleto@, instead of manually using this function (which is
-- possible but tedious), see the 'renderQueryToText' function (along with
-- 'renderQuerySelect', 'renderQueryUpdate', etc).
toRawSql
  :: (SqlSelect a r, BackendCompatible SqlBackend backend)
  => Mode -> (backend, IdentState) -> SqlQuery a -> (TLB.Builder, [PersistValue])
toRawSql mode queryEnv =
    snd . toRawSqlWith mode queryEnv

-- | Render a 'SqlQuery' into the internal representation of the returned
-- result, the text 'TLB.Builder', and the @['PersistValue']@ to interpolate
-- into the query text.
toRawSqlWith
    :: (SqlSelect a r, BackendCompatible SqlBackend backend)
    => Mode -> (backend, IdentState) -> SqlQuery a -> (a, (TLB.Builder, [PersistValue]))
toRawSqlWith mode (conn, firstIdentState) query =
    let ((ret, sd), finalIdentState) =
            flip S.runState firstIdentState $
            W.runWriterT $
            unQ query
        deleteRepeatedNewlines txt =
            let
                (preNewlines, rest) = TL.break (== '\n') txt
                (_, rest') = TL.break (/= '\n') rest
             in
                if TL.null rest'
                    then preNewlines <> "\n"
                    else preNewlines <> "\n" <> deleteRepeatedNewlines rest'

        SideData distinctClause
                 fromClauses
                 setClauses
                 whereClauses
                 groupByClause
                 havingClause
                 orderByClauses
                 limitClause
                 lockingClause
                 cteClause = sd
        -- Pass the finalIdentState (containing all identifiers
        -- that were used) to the subsequent calls.  This ensures
        -- that no name clashes will occur on subqueries that may
        -- appear on the expressions below.
        info = (projectBackend conn, finalIdentState)

        overFst f (a, b) = (f a, b)
        convert = TLB.fromLazyText . deleteRepeatedNewlines . TL.strip . TLB.toLazyText

    in (,) ret $ overFst convert $ mconcat $ intersperse ("\n", [])
        [ makeCte        info cteClause
        , makeInsertInto info mode ret
        , makeSelect     info mode distinctClause ret
        , makeFrom       info mode fromClauses
        , makeSet        info setClauses
        , makeWhere      info whereClauses
        , makeGroupBy    info groupByClause
        , makeHaving     info havingClause
        , makeOrderBy    info orderByClauses
        , makeLimit      info limitClause
        , makeLocking    info lockingClause
        ]
```

src/Database/Esqueleto/Internal/Internal.hs
ulidtko: Yep you're right; I initially tried something different, this had more constructors — but it ended up isomorphic to Maybe ().

Will refactor.

src/Database/Esqueleto/Internal/Internal.hs
ulidtko: Sure, I though about that, but `tellReturning` is in `Internal` module... its haddock says so, too:

```haskell
-- | (Internal) Remember a @RETURNING@ clause in a query
tellReturning :: ReturningClause -> SqlQuery ()
```

The public interface is supposed to be only the added functions in `Database.Esqueleto.PostgreSQL` (plus associated instances).

Further, the rest of query builder logic didn't seem like it provided any sql correctness guarantees internally.

src/Database/Esqueleto/Internal/Internal.hs
ulidtko: the most important one is the syntax fix on `sub_select` deprecation notice :smiley: (screenshot in PR head post)

ulidtko:
Thanks for review @parsonsmatt :pray:

Easy nits done; I'll work on this some more. Your suggestion does make sense, I see its merits.

One question I got which could use your input: how to implement the `INSERT … RETURNING …` variant?  I've got at the point of building the intended query, but then [hit issues](#issuecomment-1702647007) trying to combine the output conduit with a way to simultaneously *input* values into the query. Modulo the `(?)` placeholders, none of other query types allow to both input & output entire relations at the same time. So I struggled to come up with a way to do it. Any approach you can suggest?

src/Database/Esqueleto/Internal/Internal.hs
parsonsmatt: Yeah,it's not a generally safe thing, but I do want to avoid adding more unsafety where possible.

belevy:
So I just hit a reason to implement `insertSelectWithConflictReturning` and it led me to want something like `SqlQuery` specifically for inserts. My original type was `SqlInsert insertType returningType`, I ended up moving to `SqlInsert backend insertType returningType` though really it could probably just be monomorphic in `SqlBackend`. 

So my idea for would be something like 

```haskell 
data OnConflictDo ent 
  = DoNothing 
  | DoUpdate (SqlExpr (Entity ent) -> SqlExpr (Entity ent) -> [SqlExpr (Entity ent) ->SqlExpr Update])
runInsertReturning :: SqlSelect a r => SqlInsert backend insertType a -> ReaderT backend m [r]
onConflict :: EI.KnowResult uniq ~ Unique ent => uniq -> OnConflictDo ent -> SqlInsert backend ent a -> SqlInsert backend ent a

let doNothing = DoNothing
in insertFrom query 
  & onConflict UniqueField doNothing
  & runInsertReturning
```

belevy:
Having worked with this a bit more, the backend can always just be SqlBackend inside of the SqlInsert type. Additionally runInsertReturning should take a projection function to determine the output rather than figuring out how to return an `a` directly.

stuebinm:
Any possible way forward on this? I just hit a case today where I'd really like to have this, but the PR history seems discouraging …

ulidtko:
Thanks for the bump @stuebinm, I took a workaround and completely forgot about this branch 🤦 

Will review and try to resume progress.

PR #363, opened 2023-05-13
Add missing just to left join examples in the Haddocks

### LINKED ISSUE #307: Example 2: leftJoin is missing a call to just
_Small problem within the documentation:_ Example 2 in Module `Database.Esqueleto` and `Database.Esqueleto.Experimental` is missing a call to `just (`...`)` in two places, since the example shows a `leftJoin` and not an `innerJoin`, hence the left-hand side of the `on` must be wrapped with `just`. Here is my suggestion for the corrected code:
```haskell
select $
from $ \(people `LeftOuterJoin` blogPosts) -> do
on (just (people ^. PersonId) ==. blogPosts ?. BlogPostAuthorId)
where_ (people ^. PersonAge >. val 18)
pure (people, blogPosts)
```
and also the suggested fix for the new syntax
```haskell
select $ do
(people :& blogPosts) <-
    from $ table @Person
    `leftJoin` table @BlogPost
    `on` (\(people :& blogPosts) ->
            just (people ^. PersonId) ==. blogPosts ?. BlogPostAuthorId)
where_ (people ^. PersonAge >. val 18)
pure (people, blogPosts)
```
I found this quite irritating, but it is quick to fix. 

_Alternatively,_ I would suggest adding these two definitions to avoid some parenthesis and fix the issue in another way:
```haskell
infixl 4 =?.
(=?.) ::  PersistField typ => E.SqlExpr (E.Value typ) ->  E.SqlExpr (E.Value (Maybe typ)) -> E.SqlExpr (E.Value Bool)
(=?.) a b = just a ==. b

infixl 4 ?=.
(?=.) ::  PersistField typ => E.SqlExpr (E.Value (Maybe typ)) ->  E.SqlExpr (E.Value typ) -> E.SqlExpr (E.Value Bool)
(?=.) a b = a ==. just b
```
I could also provide a pull request, if so desired.

duplode:
This fixes #307, which has recently resurfaced on [a Stack Overflow question](https://stackoverflow.com/q/76234412/2751851). I believe the examples adjusted here were the only ones still missing `just`.

After submitting your PR:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,9 @@
+3.5.9.1
+=======
+- @duplode
+    - [#363](https://github.com/bitemyapp/esqueleto/pull/363)
+      - Add missing `just` to left join examples in the Haddocks
+
 3.5.9.0
 =======
 - @9999years
--- a/esqueleto.cabal
+++ b/esqueleto.cabal
@@ -2,7 +2,7 @@ cabal-version: 1.12
 
 name:           esqueleto
 
-version:        3.5.9.0
+version:        3.5.9.1
 synopsis:       Type-safe EDSL for SQL queries on persistent backends.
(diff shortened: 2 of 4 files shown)


parsonsmatt:
Released, thank you!

PR #365, opened 2023-06-08
Name conflict dodges

parsonsmatt:
This PR adds `isNothing_`, `groupBy_`, and `updateE` as "non-conflicting" variants of the other esqueleto  functions. This helps avoid name conflicts and makes it easier to import `esqueleto` with other libraries.

After submitting your PR:

### DIFF
--- a/src/Database/Esqueleto/Internal/Internal.hs
+++ b/src/Database/Esqueleto/Internal/Internal.hs
@@ -261,6 +261,13 @@ on expr = Q $ W.tell mempty { sdFromClause = [OnClause expr] }
 groupBy :: (ToSomeValues a) => a -> SqlQuery ()
 groupBy expr = Q $ W.tell mempty { sdGroupByClause = GroupBy $ toSomeValues expr }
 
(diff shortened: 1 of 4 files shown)


src/Database/Esqueleto/Internal/Internal.hs
```diff
@@ -261,6 +261,13 @@ on expr = Q $ W.tell mempty { sdFromClause = [OnClause expr] }
 groupBy :: (ToSomeValues a) => a -> SqlQuery ()
 groupBy expr = Q $ W.tell mempty { sdGroupByClause = GroupBy $ toSomeValues expr }
 
+-- | An alias for 'groupBy' that avoids conflict with the term from "Data.List"
+-- 'Data.List.groupBy'.
+--
+-- @since 3.5.10.0
+groupBy_ :: (ToSomeValues a) => a -> SqlQuery ()
+groupBy_  = groupBy
```
parsonsmatt, line 269: To me, the main question is: do we want to deprecate the other function names at some point and only have these `_` suffixes?

PR #373, opened 2023-08-14
Alias CTEs upon initialising

### LINKED ISSUE #372: Joining on a CTE multiple times creates name clashes
Only tested on version `3.5.10.1` on MySQL.

When you include a CTE in a join, that CTE is not given a fresh alias. This means that including that CTE twice (or more times) in a join chain, both instances of the CTE are given the same name, which is invalid SQL.

A small example:

```haskell
share [mkPersist sqlSettings] [persistLowerCase|
  A
    k Int
    v Int
    Primary k
  
  B
    k Int
    v Int
    Primary k
|]

q :: SqlQuery (SqlExpr (Value Int), SqlExpr (Value Int))
q = do
  bCte <- with $ do
    b <- from $ table @B
    return b
  
  a :& b1 :& b2 <- from $ table @A
    `innerJoin` bCte
    `on` (\(a :& b) -> a.k ==. b.k)
    `innerJoin` bCte
    `on` (\(a :& _ :& b2) -> a.k ==. b2.k)
  return (a.k, a.v +. b1.v +. b2.v)
```

This is the SQL code that gets generated:
```sql
WITH `cte` AS 
  SELECT `b`.`k` AS `v_k`, `b`.`v` AS `v_v`
  FROM `b`
)

SELECT `a`.`k`,  ((`a`.`v` + `cte`.`v_v`) + `cte`.`v_v`)
FROM `a`
INNER JOIN `cte`
  ON `a`.`k` = `cte`.`v_k`
INNER JOIN `cte`
  ON `a`.`k` = `cte`.`v_k`;
```

As you can see, both the first and the second join are refered to as `cte`. Ideally, Esqueleto would produce something like this
```sql
WITH `cte` AS 
  SELECT `b`.`k` AS `v_k`, `b`.`v` AS `v_v`
  FROM `b`
)

SELECT `a`.`k`,  ((`a`.`v` + `cte_1`.`v_v`) + `cte_2`.`v_v`)
FROM `a`
INNER JOIN `cte` AS `cte_1`
  ON `a`.`k` = `cte_1`.`v_k`
INNER JOIN `cte` AS `cte_2
  ON `a`.`k` = `cte_2`.`v_k`;
```

RikvanToor:
Hello. :)

This PR fixes #372 by adding a fresh alias whenever a common table expression is used within a query.

As described in the issue, previously this Haskell code
```haskell
share [mkPersist sqlSettings] [persistLowerCase|
  A
    k Int
    v Int
    Primary k
  
  B
    k Int
    v Int
    Primary k
|]

q :: SqlQuery (SqlExpr (Value Int), SqlExpr (Value Int))
q = do
  bCte <- with $ do
    b <- from $ table @B
    return b
  
  a :& b1 :& b2 <- from $ table @A
    `innerJoin` bCte
    `on` (\(a :& b) -> a.k ==. b.k)
    `innerJoin` bCte
    `on` (\(a :& _ :& b2) -> a.k ==. b2.k)
  return (a.k, a.v +. b1.v +. b2.v)
```

would generated this SQL query
```sql
WITH `cte` AS (
  SELECT `b`.`k` AS `v_k`, `b`.`v` AS `v_v`
  FROM `b`
)

SELECT `a`.`k`,  ((`a`.`v` + `cte`.`v_v`) + `cte`.`v_v`)
FROM `a`
INNER JOIN `cte`
  ON `a`.`k` = `cte`.`v_k`
INNER JOIN `cte`
  ON `a`.`k` = `cte`.`v_k`;
```

which is invalid. After this PR, it instead generates this SQL query:
```sql
WITH `cte` AS (
  SELECT `b`.`k` AS `v_k`, `b`.`v` AS `v_v`
  FROM `b`
)

SELECT `a`.`k`,  ((`a`.`v` + `cte2`.`v_v`) + `cte3`.`v_v`)
FROM `a`
INNER JOIN `cte` AS `cte2`
  ON `a`.`k` = `cte2`.`v_k`
INNER JOIN `cte` AS `cte3`
  ON `a`.`k` = `cte3`.`v_k`
```

I'm not 100% satisfied with this, as the aliasing is only required when a single CTE is used multiple times in a single query, but my PR applies the aliasing at all times. I was not sure how to check for such a situation from within the context of the `with` function, so perhaps this PR can serve as a conversation about that. Thanks a lot!

---

After submitting your PR:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,9 @@
+3.5.10.2
+========
+- @RikvanToor
+    - [#373](https://github.com/bitemyapp/esqueleto/pull/373)
+        - Fix name clashes when using CTEs multiple times
+
 3.5.10.1
 ========
 - @9999years
--- a/esqueleto.cabal
+++ b/esqueleto.cabal
@@ -2,7 +2,7 @@ cabal-version: 1.12
 
 name:           esqueleto
 
-version:        3.5.10.1
+version:        3.5.10.2
 synopsis:       Type-safe EDSL for SQL queries on persistent backends.
 description:    @esqueleto@ is a bare bones, type-safe EDSL for SQL queries that works with unmodified @persistent@ SQL backends.  Its language closely resembles SQL, so you don't have to learn new concepts, just new syntax, and it's fairly easy to predict the generated SQL and optimize it for your backend. Most kinds of errors committed when writing SQL are caught as compile-time errors---although it is possible to write type-checked @esqueleto@ queries that fail at runtime.
                 .
--- a/src/Database/Esqueleto/Experimental/From/CommonTableExpression.hs
+++ b/src/Database/Esqueleto/Experimental/From/CommonTableExpression.hs
@@ -53,7 +53,10 @@ with query = do
     let clause = CommonTableExpressionClause NormalCommonTableExpression ident (\info -> toRawSql SELECT info aliasedQuery)
     Q $ W.tell mempty{sdCteClause = [clause]}
     ref <- toAliasReference ident aliasedValue
-    pure $ From $ pure (ref, (\_ info -> (useIdent info ident, mempty)))
+    pure $ From $ do
+      newIdent <- newIdentFor (DBName "cte")
+      localRef <- toAliasReference newIdent ref
+      pure (localRef, (\_ info -> (useIdent info ident <> " AS " <> useIdent info newIdent, mempty)))
 
 -- | @WITH@ @RECURSIVE@ allows one to make a recursive subquery, which can
 -- reference itself. Like @WITH@, this is supported in most modern SQL engines.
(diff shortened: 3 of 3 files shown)


belevy:
please add a test that would fail to work without this fix. As far as aliasing only on duplicates, how do we handle it with regular table joins?

jonathanmoregard:
> please add a test that would fail to work without this fix. As far as aliasing only on duplicates, how do we handle it with regular table joins?

See https://github.com/bitemyapp/esqueleto/issues/372.

I'm pretty sure I've run into the same issue in our production app :)

parsonsmatt:
I'll incorporate the test and try to get this released today. Thanks!

parsonsmatt:
Closing in favor of #410 - thanks for the great work!

RikvanToor:
@parsonsmatt Thanks a lot! This has been in the back of my mind for a very long time, but I never got around to doing it.

PR #370, opened 2023-08-04
Generate `ToMaybe` in `deriveEsqueletoRecord`

9999years:
Before submitting your PR, check that you've:

- [ ] Bumped the version number.
- [ ] Documented new APIs with [Haddock markup](https://www.haskell.org/haddock/doc/html/index.html).
- [ ] Added [`@since` declarations](http://haskell-haddock.readthedocs.io/en/latest/markup.html#since) to the Haddock.
- [ ] Ran `stylish-haskell` and otherwise adhered to the [style guide](https://github.com/bitemyapp/esqueleto/blob/master/style-guide.yaml).

After submitting your PR:

- [ ] Update the Changelog.md file with a link to your PR.
- [ ] Check that CI passes (or if it fails, for reasons unrelated to your change, like CI timeouts).

PR #378, opened 2023-09-21
Generate `ToMaybe` in `deriveEsqueletoRecord`

halogenandtoast:
replaces #370

After submitting your PR:

### DIFF
--- a/src/Database/Esqueleto/Record.hs
+++ b/src/Database/Esqueleto/Record.hs
@@ -757,7 +757,11 @@ toMaybeTDec RecordInfo {..} = do
   let binders = Nothing
       lhs = (ConT ''ToMaybeT) `AppT` (ConT sqlName)
       rhs = ConT sqlMaybeName
+#if MIN_VERSION_template_haskell(2,15,0)
   pure $ TySynInstD $ TySynEqn binders lhs rhs
+#else
+  pure $ TySynInstD $ TySynEqn lhs rhs
+#endif
 
 -- | Generates a `toMaybe value = ...` declaration for the given record.
 toMaybeDec :: RecordInfo -> Q Dec
--- a/src/Database/Esqueleto/Record.hs
+++ b/src/Database/Esqueleto/Record.hs
@@ -130,11 +130,21 @@ data DeriveEsqueletoRecordSettings = DeriveEsqueletoRecordSettings
     -- name to produce the SQL record's type name and constructor name.
     --
     -- @since 3.5.8.0
+  , sqlMaybeNameModifier :: String -> String
+    -- ^ Function applied to the Haskell record's type name and constructor
+    -- name to produce the 'ToMaybe' record's type name and constructor name.
+    --
+    -- @since 3.5.11.0
   , sqlFieldModifier :: String -> String
     -- ^ Function applied to the Haskell record's field names to produce the
     -- SQL record's field names.
     --
     -- @since 3.5.8.0
+  , sqlMaybeFieldModifier :: String -> String
+    -- ^ Function applied to the Haskell record's field names to produce the
+    -- 'ToMaybe' SQL record's field names.
+    --
+    -- @since 3.5.11.0
   }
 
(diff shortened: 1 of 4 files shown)


halogenandtoast:
Opening this up for conversation since the draft didn't get eyes and I think this is nearly there.

test/Common/Record.hs
```diff
@@ -8,25 +8,34 @@
 {-# LANGUAGE MultiParamTypeClasses #-}
 {-# LANGUAGE NamedFieldPuns #-}
 {-# LANGUAGE OverloadedLabels #-}
+{-# LANGUAGE OverloadedRecordDot #-}
```
parsonsmatt, line 11: We probably need to gate this behind CPP for the GHC version since we support GHC that doesn't have this feature yet

parsonsmatt (commented):
Could you write a test that shows it correctly working with `Nothing` values? ie where the DB gives us `NULL` back?

halogenandtoast:
@parsonsmatt did you mean something different from: https://github.com/bitemyapp/esqueleto/pull/378/files#diff-bef0a2133965554f3b49d699a015a29f235548c9ae5edc52ce80cd348332720dR282-R300

test/Common/Record.hs
halogenandtoast: I assume gating the whole file makes sense here?

halogenandtoast:
Confirmed this compiles and all existing test run and pass on 8.10.7

test/Common/Record.hs
```diff
@@ -235,3 +260,92 @@ testDeriveEsqueletoRecord = describe "deriveEsqueletoRecord" $ do
...
+                 _ -> True)
+        liftIO $ sortedRecords !! 1
+          `shouldSatisfy`
+          (\case (_ :& _ :& Just (MyNestedRecord {myRecord = MyRecord {myName = "Rebecca", myAddress = Nothing}})) -> True
+                 _ -> False)
+
+#else
+    it "is only supported in GHC 9.2 or above" $ \_ -> do
+        pending
+#endif
```
parsonsmatt, line 351: The feature is supported in GHC <9.2, but you do need to use `getField @"blah"` explicitly instead of `(.blah)`. So we need to gate *use* of `OverloadedRecordDot` behind CPP to support older GHC, but not the entire feature/tests.

Actually, it is probably easier to just `getField` directly? Then there's no CPP necessary

halogenandtoast:
I don't have a machine I can run 8.6.5 on until November 4th so fixing this will have to wait until then unless someone else can do the update here.

src/Database/Esqueleto/Record.hs
```diff
@@ -652,3 +728,231 @@ toAliasReferenceDec RecordInfo {..} = do
...
+toMaybeTDec :: RecordInfo -> Q Dec
+toMaybeTDec RecordInfo {..} = do
+  let binders = Nothing
+      lhs = (ConT ''ToMaybeT) `AppT` (ConT sqlName)
+      rhs = ConT sqlMaybeName
+#if MIN_VERSION_template_haskell(2,15,0)
+  pure $ TySynInstD $ TySynEqn binders lhs rhs
+#else
+  pure $ TySynInstD sqlName $ TySynEqn [lhs] rhs
+#endif
```
parsonsmatt, line 764:
```suggestion
-- | Generates a `type ToMaybeT ... = ...` declaration for the given record.
toMaybeTDec :: RecordInfo -> Q Dec
toMaybeTDec RecordInfo {..} = do
  pure $ mkTySynInstD ''ToMaybeT (ConT sqlName) (ConT sqlMaybeName)
  where
    mkTySynInstD className lhsArg rhs =
#if MIN_VERSION_template_haskell(2,15,0)
        let binders = Nothing
            lhs = ConT className `AppT` lhsArg
        in
            TySynInstD $ TySynEqn binders lhs rhs
#else
       TySynInstD className $ TySynEqn [lhsArg] rhs
#endif
```

PR #379, opened 2023-09-29
Fix semantics of `not_` applied over `&&.` and `||.`

### LINKED ISSUE #359: `not_` is broken
```haskell
share [mkPersist sqlSettings] [persistLowerCase|
  Person
    name String
    age Int
    deriving Eq Show
|]

query :: SqlPersistM [Entity Person]
query = select $ from \(person :: SqlExpr (Entity Person)) -> do
  where_ $ not_ (person.name !=. val "foo" &&. person.age >. val @Int 15)
  pure person
```
produces
```sql
SELECT "person"."id", "person"."name", "person"."age" FROM "person" WHERE NOT ("person"."name" != ?) AND ("person"."age" > ?); 
```
expected
```sql
SELECT "person"."id", "person"."name", "person"."age" FROM "person" WHERE NOT (("person"."name" != ?) AND ("person"."age" > ?)); 
```

blujupiter32:
This PR fixes #359 by always wrapping arguments to `not_` in parentheses when building SQL queries. Some tests combining `not_` with `&&.` and `||.` have also been added to `Common.Test`.

This is technically a breaking change: `not_ (x &&. y)` is no longer equivalent to `(not_ x) &&. y`, for example, although the equivalence is probably unintended.

---

  - `stylish-haskell` fails to parse the `DEPRECATED` pragma for `sub_select` in `Database.Esqueleto.Internal.Internal`.

After submitting your PR:

### DIFF
--- a/test/Common/Test.hs
+++ b/test/Common/Test.hs
@@ -912,6 +912,13 @@ testSelectWhere = describe "select where_" $ do
                 asserting $ ret `shouldBe` [()]
 
     describe "when using not_" $ do
+        itDb "works for a single expression" $ do
+            ret <-
+                select $
+                pure $ not_ $ val True
+            asserting $ do
+                ret `shouldBe` [Value False]
+
         itDb "works for a simple example with (>.) [uses just . val]" $ do
             _   <- insert' p1
             _   <- insert' p2
--- a/test/Common/Test.hs
+++ b/test/Common/Test.hs
@@ -878,16 +878,6 @@ testSelectWhere = describe "select where_" $ do
                return p
         asserting $ ret `shouldBe` [ p1e ]
 
-    itDb "works for a simple example with (>.) and not_ [uses just . val]" $ do
-        _   <- insert' p1
-        _   <- insert' p2
-        p3e <- insert' p3
-        ret <- select $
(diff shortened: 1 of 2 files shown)


parsonsmatt:
This is great work, thank you! I apologize for not merging and releasing earlier. I believe this should be a patch version bump, so I will release asap.

test/Common/Test.hs
```diff
@@ -921,6 +911,44 @@ testSelectWhere = describe "select where_" $ do
                         , val $ PointKey 5 6 )
                 asserting $ ret `shouldBe` [()]
 
+    describe "when using not_" $ do
```
parsonsmatt, line 914:
```suggestion
    describe "when using not_" $ do
        itDb "works for a single expression" $ do
            ret <-
                select $
                pure $ not_ $ val True
            asserting $ do
                ret `shouldBe` [Value False]

```

parsonsmatt: wanted to add one more test for the solo case

parsonsmatt (approved):

parsonsmatt:
Tested this on the work repo, only one failure relating to a golden test. Thanks!

PR #394, opened 2024-06-07
Use TH quotes more to eliminate some CPP

TeofilC:
This eliminates some CPP by switching to using TH quotes.

After submitting your PR:

### DIFF
--- a/src/Database/Esqueleto/Record.hs
+++ b/src/Database/Esqueleto/Record.hs
@@ -410,7 +410,7 @@ sqlSelectColsDec RecordInfo {..} = do
              in foldl' helper (VarE f1) rest
 
   identInfo <- newName "identInfo"
-  [d| sqlSelectCols $(varP identInfo) $(pure $ RecP sqlName fieldPatterns) =
+  [d| $(varP 'sqlSelectCols) = \ $(varP identInfo) $(pure $ RecP sqlName fieldPatterns) ->
         sqlSelectCols $(varE identInfo) $(pure joinedFields)
     |]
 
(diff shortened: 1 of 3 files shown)


TeofilC:
What do you think about this @parsonsmatt ? If you are a fan of this direction, I can try to also get rid of the CPP caused by `DoE`

TeofilC:
I'll investigate these failures soon

src/Database/Esqueleto/Record.hs
```diff
@@ -951,23 +880,7 @@ sqlMaybeSelectColCountDec RecordInfo {..} = do
...
-              AppE (VarE 'sqlSelectColCount) $
-                ParensE $
-                  AppTypeE
-                    (ConE 'Proxy)
-                    joinedTypes
-          )
-          -- `where` clause.
-          []
-      ]
+  [d| sqlSelectColCount _ = sqlSelectColCount (Proxy @($(pure joinedTypes))) |]
```
parsonsmatt, line 883: Huh - there are two mentions of `sqlSelectColCount`. This is one. Is it posible that the `[d| ... |]` here is doing `newName` instead of using the name in scope?

TeofilC: I think you are right. This is quite odd behaviour (but also the type of issues I was hoping to find with this PR). I'll make a GHC ticket.

Strangely as welll
`[d| $(varP 'foo) = ...|]` is fine but as soon as I try to add arguments, it fails to parse, so I moved those into a lambda

TeofilC: GHC issue: https://gitlab.haskell.org/ghc/ghc/-/issues/25435

parsonsmatt (approved):
This looks much better, thank you!

TeofilC:
Cheers! Happy holidays

PR #406, opened 2024-11-26
Do not error on any AppT in sqlMaybeSelectProcessRowDec

csamak:
No need to be this restrictive in deriveEsqueletoRecord For example, AppT ListT _ and AppT (ConT _) _ are both fine but would throw an error here before this change.

After submitting your PR:

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,4 @@
-3.5.14.0
+3.5.13.1
 ========
 - @csamak
     - [#405](https://github.com/bitemyapp/esqueleto/pull/405)
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,9 @@
+3.5.14.0
+========
+- @csamak
+    - [#405](https://github.com/bitemyapp/esqueleto/pull/405)
+      - Fix a bug introduced in 3.5.12.0 where deriveEsqueletoRecord incorrectly errors
+
 3.5.13.0
 ========
 - @ac251
--- a/esqueleto.cabal
+++ b/esqueleto.cabal
@@ -2,7 +2,7 @@ cabal-version: 1.12
 
 name:           esqueleto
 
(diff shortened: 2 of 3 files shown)


src/Database/Esqueleto/Record.hs
```diff
@@ -935,9 +935,9 @@ sqlMaybeSelectProcessRowDec RecordInfo {..} = do
       ]
   where
     wrapJust x = case x of
-      AppT (ConT ((==) ''Entity -> True)) _innerType -> id
       ((ConT ((==) ''Maybe -> True)) `AppT` _inner) -> AppE (ConE 'Just)
       (ConT _) -> id
+      (AppT _ _) -> id
       _ -> error $ show x
```
parsonsmatt, line 941: Do we want to keep the `error` here? I feel like we may want to just have `id` as the fallback case instead of erroring on unknown constructors

parsonsmatt (approved):
hell yeah! Let's make a patch version bump + changelog entry and I'll get this released + deprecate the Hackage versions.

changelog.md
```diff
@@ -1,3 +1,9 @@
+3.5.14.0
```
parsonsmatt, line 1: Since this is a bugfix we can do a patch release
```suggestion
3.5.13.1
```

esqueleto.cabal
```diff
@@ -2,7 +2,7 @@ cabal-version: 1.12
 
 name:           esqueleto
 
-version:        3.5.13.0
+version:        3.5.14.0
```
parsonsmatt, line 5:
```suggestion
version:        3.5.13.1
```

parsonsmatt (approved):

parsonsmatt:
Tested on the work codebase, all good! Uploaded as 3.5.13.1

